PR #49, opened 2014-01-28
CRM-14126 - Wrap CURLOPT_FOLLOWLOCATION in a check for open_basedir and safe_mode

lolas-freeform:
---
- CRM-14126: HttpClient.php library used by CiviCRM issues a PHP warning when open_basedir is set
  http://issues.civicrm.org/jira/browse/CRM-14126


colemanw:
Please submit against master branch.

PR #66, opened 2014-06-07
CRM-14814 CRM-8744 breaks on SMTP svrs requiring secure authentication

kenwest:
See forum discussion http://forum.civicrm.org/index.php/topic,21960

The change made for CRM-8744 prevents CiviCRM connecting to a SMTP server unless that server provides authentication in the clear. This is insecure.

Notes:
1) The latest version of http://pear.php.net/package/Net_SMTP doesn't include this patch (released 2013-07-05) which tells me it's not an important patch
2) The source of the patch http://www.pear-forum.org/post-4935.html has gone off-air so we can't see why this patch was made or assess how well it was tested etc
3) I can demonstrate (in the Forum discussion) that this patch breaks the code
4) This patch creates a security issue, because it forces people to authenticate SMTP in the clear

---

 * [CRM-14814: CRM-8744 breaks on SMTP servers requiring secure TLS authentication](https://issues.civicrm.org/jira/browse/CRM-14814)
 * [CRM-8744: Upgrade and patch smtp library](https://issues.civicrm.org/jira/browse/CRM-8744)

### DIFF
--- a/Net/SMTP.php
+++ b/Net/SMTP.php
@@ -601,8 +601,6 @@ class Net_SMTP
          * (SSL) socket connection. */
         if ($tls && version_compare(PHP_VERSION, '5.1.0', '>=') &&
             extension_loaded('openssl') && isset($this->_esmtp['STARTTLS']) &&
-            // CiviCRM: CRM-8744
-            ($this->_esmtp['STARTTLS'] == true) &&
             strncasecmp($this->host, 'ssl://', 6) !== 0) {
             /* Start the TLS connection attempt. */
             if (PEAR::isError($result = $this->_put('STARTTLS'))) {


totten:
+1. Merged. Feedback on forums was positive, and this PR aligns us better with upstream's code.

It's not clear to me if this will undo whatever benefit/fix was originally provided by CRM-8744 / http://www.pear-forum.org/post-4935.html -- or if  `post-4935.html` was just Really Bad(tm) to begin with. Worst case... it trades one bug for another, but it's no longer our bug... it would be a bug in the upstream library. If that happens, we should either refer the problem upstream or switch to swiftmailer.org.

Aside: It looks like [upstream](https://github.com/pear/Net_SMTP/blob/master/Net/SMTP.php) has made some other TLS-related fixes and improved the syntax (boo PHP-4, yay PHP-7), so I might send another PR to upgrade.

PR #75, opened 2014-10-01
update logic to allow civicrm to be installed in /profiles

joelcollinsdc:
https://issues.civicrm.org/jira/browse/CRM-15405


colemanw:
Resubmitted as https://github.com/civicrm/civicrm-packages/pull/95

PR #88, opened 2015-01-07
CRM-15789 - Remove civicrm.css styles redundant with jQuery UI

colemanw:


colemanw:
Should be merged in tandem with https://github.com/civicrm/civicrm-core/pull/4861

PR #112, opened 2015-05-08
PHPUnit: Fix assertTableRowCount()

antrik:
This method can't be static, as it accesses $this->getConnection().

This fix has been upstream for years (
https://github.com/sebastianbergmann/dbunit/commit/fce1b25ba4070eebf860653ec9f5b222c2a018ba
) -- but CiviCRM prefers sticking with an ancient buggy version of
PHPUnit/DBUnit...


totten:
The test scripts were revised to use standalone PHPUnit (rather the ancient/buggy version in packages.git)!

PR #114, opened 2015-07-02
un-minify "jquery.ui.datepicker.validation.pack.js"

onlyjob:
CRM-16426: replaced "jquery.ui.datepicker.validation.pack.js" with its
 original uncompressed version; added compatibility symlink.

Signed-off-by: Dmitry Smirnov <onlyjob@member.fsf.org>


totten:
Agree we should have a copy of the pristine file.

Symlinks worry me -- e.g. they're awkward with Windows (grrr, Windows) and with .zip files (like `civicrm-*-wordpress.zip` and `civicrm-*-joomla.zip`). It's safer to copy in both the pristine and minified versions (as done with several other jquery plugins).

FWIW, if we could call the file `jquery.ui.datepicker.validation.min.js`, then Civi would automatically switch between minified and pristine versions (depending on whether debugging is enabled).

onlyjob:
My intention is to completely eliminate minified file.
Once another commit update name of this file in `CRM/Core/Resources.php` I was planning to drop compatibility symlink. However I'm not sure how you guys synchronise commits between repositories... Now I've replaced symlink with the copy of the original file. Thanks for your comments, Tim.

totten:
Ah, ok. That makes sense.

To coordinate changes across the repos, we do it the lo-tech way -- if there's a dependency between PR's in different repos, then just make a comment and include a hyperlink to the other PR. We can often merge them concurrently. (This is a bit easier with the JS code than with the PHP code because Jenkins' doesn't have much test-coverage over the JS code. But regardless, the technique you describe sounds pretty conscientious and should be safe.)

colemanw:
I'm really confused about this change. Our practice is to include minified and non-minified js side-by-side in packages. What you've done here doesn't accomplish that, nor does it provide more complete license info than what was there already.

I'd propose we restore the minified version and keep the uncompressed version around as well. Renaming the file to end with .min.js instead of .pack.js per Tim's suggestion would be good for consistency and to allow the non-minified version to be used appropriately.

colemanw:
This sorts it: https://github.com/civicrm/civicrm-packages/pull/115 https://github.com/civicrm/civicrm-core/pull/6125

@onlyjob to clarify our practice and how it works: providing minified and raw js files side-by-side makes the source code accessible without losing the performance benefits of using minified code in production. When debug-mode is enabled, Civi will automatically look for the non-minified version and supply that to the browser, which is helpful for developers and anyone else interested in seeing the full source code.

onlyjob:
Thanks for taking care of renaming the file to be more consistent.

@colemanw, I have a feeling that you do not understand the problem I was trying to address: minified file without source is non-free and should not be distributed. When original files are accompanied by minified files, it is just a bad and unnecessary practice (similar to committing compiled C/C++ binaries). Ideally minified files should be produced from source on build time (if enabled with optional argument to build system). In Debian we have to remove all pre-built binaries or re-build 'em from source. It is quite time consuming to remove minified files from released archive of CiviCRM.

Besides personally I dislike minified javascript files. They are unnecessary and they add little to performance of properly configured web server (assuming gzip compression is enabled). They complicate debugging and hide logic from user. On few occasions I've seen minified files exhibiting problems not manifesting in the original source. I'm also thinking about parsers -- although I did not do any client-side (browser) performance study I know for a fact that some XML parsers are working slower with "minified" XML which is just one very long line without breaks. If same is true for browsers then minified files can actually make user experience worse. IMHO they just do not worth the effort and if I want to I can manually minify all javascripts with simple command in few minutes only. I think it is the best to entirely drop all minified files and go ahead with their original sources only.

PR #116, opened 2015-07-08
CRM-16387 - Backport Symfony dependencies

totten:


totten:
This will be required for the other test to pass, and it shouldn't pose any danger by itself because it only imports new classes.

PR #124, opened 2015-08-14
Move ezc components from packages to composer.json

totten:
This is in anticipation of proper forking and applying CRM-15499.

---

 * [CRM-15499: courier imap server responds "max atom size too small" if more than 5,000 bounce messages exist](https://issues.civicrm.org/jira/browse/CRM-15499)


totten:
Note: This depends on https://github.com/civicrm/civicrm-core/pull/6494

PR #129, opened 2015-09-17
CRM-16790 - Remove unused files from packages

colemanw:


totten:
-28,467 lines. Hot damn!

PR #135, opened 2015-12-02
Net_SMTP, Auth_SASL, Net_Socket - Remove. Will use composer.

totten:
Pre-requisite: https://github.com/civicrm/civicrm-core/pull/7361


totten:
* `Net_Socket` was a clean copy of 1.0.9. 
* I couldn't tell authoritatively if we had patches on `Auth_SASL`, but there was nothing to suggest we did. (Nothing in `VERSIONS` or `_ORIGINAL_`)
* After #66, `Net_SMTP` is better aligned with upstream, but there were a couple diffs:
   * The `Auth_SASL` calls used an older notation. This was necessary because `Auth_SASL` was out-of-date. But these PR's update both.
   * One of the error messages was modified with Civi-specific instructions ( https://github.com/civicrm/civicrm-packages/pull/135/files#diff-205d279ce3a50e2e2cfabaad983cac79L374 ). It seems like a decent idea to display more specific errors, but patching this file isn't a good way to do it. Better to patch `civicrm-core` to catch the error and report something better.

seamuslee001:
@totten i think the only thing to do here is to clean up VERSIONS.php

totten:
Replaced by https://github.com/civicrm/civicrm-packages/pull/186

PR #139, opened 2016-01-21
Remove PHPUnit

totten:
After https://github.com/civicrm/civicrm-core/pull/7624 , we'll be running the standalone version of PHPUnit.


totten:
Let's merge in this in a couple days. We should allow a little time for active PRs to clear through.

totten:
OK, we've had a fairly long transition window for the changes to `master` to propagate. We should no longer need this copy of PHPUnit.

PR #141, opened 2016-02-24
CRM-18093 extend information for query debuggers

eileenmcnaughton:
* [CRM-18093: Provide speed information on queries when CIVICRM_DEBUG_LOG_QUERY is on](https://issues.civicrm.org/jira/browse/CRM-18093)

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -2486,7 +2486,8 @@ class DB_DataObject extends DB_DataObject_Overload
 
         if (!empty($_DB_DATAOBJECT['CONFIG']['debug']) || defined('CIVICRM_DEBUG_LOG_QUERY')) {
           $timeTaken = sprintf("%0.6f", microtime(TRUE) - $time);
-          $message =  "QUERY DONE IN $timeTaken  seconds.";
+          $alertLevel = $this->getAlertLevel($timeTaken);
+          $message =  "$alertLevel QUERY DONE IN $timeTaken  seconds.";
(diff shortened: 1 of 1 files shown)


DB/DataObject.php
```diff
@@ -4278,7 +4304,30 @@ function free()
...
+    if ($timeTaken >= 20) {
+      return '****';
+    }
+    if ($timeTaken > 10) {
+      return '***';
+    }
+    if ($timeTaken > 5) {
+      return '**';
+    }
+    if ($timeTaken > 1) {
```
eileenmcnaughton: I think this is good to merge. My last thought is that * is actually not a great thing to grep for - as it is found in SELECT *

I thought about changing to 
~  1 sec
~~ 5 sec
~~~ 10 sec
~~~~ 20 sec 

But I'm on the fence cos it looks wierd & doesn't scream 'something is wrong'

totten: It's neat how you could search for "as-slow-or-slower-than" (e.g. searching for `**` will match `**` or `***`; similarly, `~~` matches `~~` or `~~~`), although it is a little cryptic. Maybe something more textual would do:

```
001 sec (slow)
005 sec (very-slow)
010 sec (very-very-slow)
020 sec (very-very-very-slow)
```

But I'm just bikeshedding.

eileenmcnaughton: hmm - that'll do! I'll make that change & then merge unless you have any further thoughts

PR #143, opened 2016-02-25
CRM-18098. Upgrade TCPDF to 6.2.12.

xurizaemon:
Select *either* this PR, or #144. This one upgrades TCPDF in packages.

---

 * [CRM-18098](https://issues.civicrm.org/jira/browse/CRM-18098)

PR #144, opened 2016-02-25
CRM-18098. Remove TCPDF, use via Composer instead.

xurizaemon:
Select this PR, or #143. This one removes TCPDF from packages, and goes with civicrm/civicrm-core#7868 which uses Composer.

---

 * [CRM-18098](https://issues.civicrm.org/jira/browse/CRM-18098)


totten:
@xurizaemon , it looks like we also need to update `composer.lock`. You can do this by running:

```bash
composer update tecnickcom/tcpdf
```

totten:
For posterity, I misfiled that comment above, but Chris caught it anyway and patched the right place (https://github.com/civicrm/civicrm-core/pull/7868).

PR #147, opened 2016-04-22
CRM-17789 - PHP7 compat for DB_DataObject

mfb:
Resolves PHP7 syntax error, unexpected 'clone' in eval()'d code.

---


totten:
Thanks, @mfb!

Change sounds sensible -- i.e. PHP7 should use PHP5 behavior instead of PHP4 behavior.

PR #148, opened 2016-05-17
Remove _ORIGINAL_ directory

rajbdilip:


eileenmcnaughton:
@totten Can I, can I?

totten:
This would be wonderful, but it's not safe. :( Basically, this is the main way we have to determine whether any packages have been forked. (Which, of course, shouldn't be done... but has been done in the past...)

The way to clean this up is incrementally -- e.g. examine each dependency in turn. *If it's safe to un-fork (to align with upstream's current release)*, then remove the dependency completely from `civicrm-packages.git`. Then declare dependency in `civicrm-core`'s `composer.json` / `composer.lock`. Ensure that the dependency is properly loaded/used from the new location.

JKingsnorth:
+0 −23,972

;]

PR #149, opened 2016-05-18
CRM-17789 free result fixes

mollux:
These fixes make it possible to for CRM_Core_DAO->freeResult() to be mysql(i) extension independent, and also fixes the DB_mysqli->freeResult to actually clear it.

This is needed for supporting PHP7 without php_mysql extension, so it doesn't break (existing) installations.

---

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -4276,18 +4276,12 @@ class DB_DataObject extends DB_DataObject_Overload
     {
         global $_DB_DATAOBJECT;
 
-        if (isset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid])) {
-            if ( is_resource( $_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]->result ) ) {
-                mysql_free_result( $_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]->result );
-            }
-            unset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]);
-        }
-
         if (isset($_DB_DATAOBJECT['RESULTFIELDS'][$this->_DB_resultid])) {
             unset($_DB_DATAOBJECT['RESULTFIELDS'][$this->_DB_resultid]);
         }
         if (isset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid])) {
             unset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]);
+            $_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]->free();
         }
         // clear the staticGet cache as well.
         $this->_clear_cache();
--- a/DB.php
+++ b/DB.php
@@ -1394,12 +1394,9 @@ class DB_result
     function free()
     {
         $err = $this->dbh->freeResult($this->result);
-        if (DB::isError($err)) {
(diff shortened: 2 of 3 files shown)


mollux:
After some testing, this PR seems to make no difference in memory usage or functionality (despite the fact that mysql_free_result is never called when using a mysqli:// dsn, it doesn't make any difference in memory usage when running the tests). I will do some more testing to be sure this PR is obsolete

DB/DataObject.php
```diff
@@ -4276,18 +4276,12 @@ function free()
...
-            }
-            unset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]);
-        }
-
         if (isset($_DB_DATAOBJECT['RESULTFIELDS'][$this->_DB_resultid])) {
             unset($_DB_DATAOBJECT['RESULTFIELDS'][$this->_DB_resultid]);
         }
         if (isset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid])) {
             unset($_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]);
+            $_DB_DATAOBJECT['RESULTS'][$this->_DB_resultid]->free();
```
totten: Wasn't this just unset in the line before?

JoeMurray:
@nielosz would you be interested in reviewing this PR?

nielosz:
Does civi's core team provide a php 7 civicrm playground?

JoeMurray:
A variety of issues remain before master branch for 4.7 will work on PHP7. But I take it you are asking whether buildkit or some other infrastructure more along the lines of a Docker container or a Vagrant image has been setup for testing of PHP7. @nganivet @totten any suggesions here? See https://github.com/djcf/civibuildkit-docker/issues/1

nganivet:
@JoeMurray I imagine it would be quite easy to modify the Vagrant buildkit to include PHP7. Here the most obvious places where a change would be needed (change php5 package name to php7):
https://github.com/civicrm/civicrm-buildkit-vagrant/blob/master/provision/provision.sh#L34
https://github.com/civicrm/civicrm-buildkit-vagrant/blob/master/provision/provision.sh#L281
https://github.com/civicrm/civicrm-buildkit-vagrant/blob/master/provision/provision.sh#L349

Then you would have to check if any changes/patches are needed for composer, PHPUnit, mailcatcher and all other PHP-based utilities used by buildkit (and buildkit itself BTW). There also are PHPMyAdmin, PHPMemchachedAdmin, OpcacheStatus, PHP_CodeSniffer and a few others, but these are not needed for running buildkit or CiviCRM so less critical.

That's all I see so far. So probably 1/2 day of work to get this done.

JoeMurray:
This seems like a useful version to build since we will be moving to support PHP7 in the next few months. Let's see if djcf is interested in creating an appropriate Docker container for us. If not, might you have the time this week?

nganivet:
Sorry Joe, I am not interested in PHP 7 anytime soon so will have to 
pass.

------ Original Message ------

DB/DataObject.php
nielosz: Agree. Thus it won't build stating "Error: Call to a member function free() on a non-object"

totten:
This has been resubmitted by Seamus in #169 to address the reported error.

PR #150, opened 2016-05-25
CRM-17789 - Support PHP 7

mollux:

### DIFF
--- a/Archive/Tar.php
+++ b/Archive/Tar.php
@@ -101,9 +101,9 @@ class Archive_Tar extends PEAR
     *                   boolean value 'true' means 'gz'.
     * @access public
     */
-    function Archive_Tar($p_tarname, $p_compress = null)
+    function __construct($p_tarname, $p_compress = null)
     {
-        $this->PEAR();
+        parent::__construct();
         $this->_compress = false;
         $this->_compress_type = 'none';
         if (($p_compress === null) || ($p_compress == '')) {
--- a/Console/Table.php
+++ b/Console/Table.php
@@ -169,7 +169,7 @@ class Console_Table
      *                         extension.
      * @param boolean $color   Whether the data contains ansi color codes.
      */
-    function Console_Table($align = CONSOLE_TABLE_ALIGN_LEFT,
+    function __construct($align = CONSOLE_TABLE_ALIGN_LEFT,
                            $border = CONSOLE_TABLE_BORDER_ASCII, $padding = 1,
                            $charset = null, $color = false)
     {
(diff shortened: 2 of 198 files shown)


totten:
@mollux when figuring out how to upgrade or un-fork packages, we often look in `_ORIGINAL_` (which is a terrible way to handle it... but it fits into the workflow).

Since there's a large batch of files being changed here, we should technically copy over the originals... or at a minimum, perhaps you could put a note in `_ORIGINAL_/README.md` or somesuch explaining that a bulk change was required for PHP7 support of the constructor/parent references.

PalanteJon:
@mollux Are you coming to CiviCon next week?  I'd like to add myself to the list of people who would like to buy you a drink.

mollux:
@PalanteJon unfortunately not, I live in Europe, but thanks! :)

mollux:
@totten the _ORIGINAL_ is indeed "interesting"... (I always delete that folder during development, as searching for methods or classes gives a lot of doubles). 
I will add a README.md to explain the big change.

totten:
A couple notes:
 * This PR has a hard dependency on https://github.com/civicrm/civicrm-core/pull/8455 , so merge 8455 before (or concurren with) 150.
 * Since the test suite doesn't run on `civicrm-packages` automatically, I'm doing a manual test on `test-ubu1204-5`. (Specifically, ` civibuild create pkgs-150-1 --patch https://github.com/civicrm/civicrm-core/pull/8455 --patch https://github.com/civicrm/civicrm-packages/pull/150` and then calling each of the major test suites.)

totten:
Results from the manual run on `test-ubu1204-5`:
 * `Civi\AllTests --exclude-group ornery`: Pass
 * `api_v3_AllTests --exclude-group ornery`: Pass
 * `CRM_AllTests  --exclude-group ornery`: Multiple failures in `CRM_Core_TransactionTest` with message `mysqli_insert_id() expects parameter 1 to be mysqli, resource given`
 * `civicrm-upgrade-test` (4.2.0*, 4.4.0*): Pass

mollux:
I will look in those failing tests this week, so we can move this forward

totten:
Taking a quick skim of https://github.com/civicrm/civicrm-core/pull/8455/commits/90c36e26d57a28d5e3a7e95e938e969177fcda38#diff-99201c02152b6d59169f6cd6c90b04c0L365 , the issue is probably that the test system on PHP 5.x still uses the `mysql` driver instead of `mysqli`.

In the context of upgrading/migrating/alternative-installers, we talked about [patching `DB::parseDSN` to always use `mysqli`](https://gist.github.com/totten/9853555a43bffffd487962fd37b05020) (even if the `CIVICRM_DSN` points to `mysql://`, treat it as `mysqli://`). I think the same patch would fix that test issue.

PR #156, opened 2016-06-09
Backport of SA-2016-09

jaapjansma:
@eileenmcnaughton could you merge this? @omarabuhussein confirmed this patch


omarabuhussein:
Yes I can confirm this patch.

eileenmcnaughton:
done! Note the tarball version of this will always be at http://dist.civicrm.org/by-date/latest/4.4/ - although we can update the release on the download page too if the time is right for that

PR #157, opened 2016-06-14
CRM-17608: Add TbsZip class [followup of pr-8474]

monishdeb:
* [CRM-17608: Merge to DOCx or ODT template](https://issues.civicrm.org/jira/browse/CRM-17608)

### DIFF
--- a/TbsZip/tbszip.php
+++ b/TbsZip/tbszip.php
@@ -1,0 +1,1003 @@
+<?php
+
+/*
+TbsZip version 2.16
+Date    : 2014-04-08
+Author  : Skrol29 (email: http://www.tinybutstrong.com/onlyyou.html)
+Licence : LGPL
+This class is independent from any other classes and has been originally created for the OpenTbs plug-in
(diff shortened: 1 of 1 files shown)


colemanw:
@monish what would you think about using the `\PhpOffice\PhpWord\Shared\ZipArchive` class instead of adding this new package?

monishdeb:
Yup I already tried but it doesn't have any function to extract and return content as text of given document file for token replacement, although you can extract it and save to new destination provided. Also as I stated earlier phpword always return/downloads blank document at the end whenever I tried to merge multiple documents into one final document. 

So for now we are using phpword here to transform document to corresponding HTML file which we later use for preview and rest are handled by this TbsZip library like
 1. returning content as a text.
 2. merging multiple contacts into one.

colemanw:
Cool

monishdeb:
Thanks :) 

So only problem is at hand is to divide final document on basis of each contact by imploding page breaks, that's still on my todo list :(

PR #158, opened 2016-06-27
CRM-19013 - Smarty - Do not truncate compiled templates

mfb:
Truncating a compiled template leads to concurrency issues if another process is including the same file. See also CRM-11189 for a description of the "bus error" encountered when one of the concurrent processes dies.

---
- [CRM-19013: Compiled templates are truncated](https://issues.civicrm.org/jira/browse/CRM-19013)
- [CRM-11189: When CiviMail renders smarty templates for mailing, it saves them to disk](https://issues.civicrm.org/jira/browse/CRM-11189)

### DIFF
--- a/Smarty/Smarty.class.php
+++ b/Smarty/Smarty.class.php
@@ -1516,40 +1516,7 @@ class Smarty
                                                   $this->_compile_id );
         $compilePath .= '.php';
 
-        //for 'string:' resource smarty might going to fail to create
-        //compile file, so make sure we should have valid path, CRM-5890
-        $matches = array( );
-        if ( preg_match( '/^(\s+)?string:/', $resource_name, $matches ) ) {
-            if ( !$this->validateCompilePath( $compilePath ) ) {
-                $compilePath = $this->_get_auto_filename( $this->compile_dir,
-                                                          time().rand(),
-                                                          $this->_compile_id );
-                $compilePath .= '.php';
-            }
-        }
-
         return $compilePath;
-    }
-
-    /**
-     *  do check can smarty create a file w/ given path.
-     */
-    function validateCompilePath( $compilePath ) {
-        //first check for directory.
-        $dirname = dirname( $compilePath );
-        if ( !is_dir( $dirname ) ) {
-            require_once(SMARTY_CORE_DIR . 'core.create_dir_structure.php');
-            smarty_core_create_dir_structure( array('dir' => $dirname ), $this );
-        }
-
-        $isValid = false;
-        if ( $fd = @fopen( $compilePath, 'wb') ) {
-            $isValid = true;
-            @fclose( $fd );
-            @unlink($compilePath);
-        }
-
-        return $isValid;
     }
 
     /**
@@ -1786,13 +1753,11 @@ class Smarty
 
         if(isset($auto_source)) {
             // make source name safe for filename
-            $_filename = urlencode(basename($auto_source));
-            $_crc32 = sprintf('%08X', crc32($auto_source));
+            $_sha256 = hash('sha256', $auto_source);
             // prepend %% to avoid name conflicts with
             // with $params['auto_id'] names
-            $_crc32 = substr($_crc32, 0, 2) . $_compile_dir_sep .
-                      substr($_crc32, 0, 3) . $_compile_dir_sep . $_crc32;
-            $_return .= '%%' . $_crc32 . '%%' . $_filename;
+            $_return .= '%%' . substr($_sha256, 0, 2) . $_compile_dir_sep .
+                        substr($_sha256, 0, 3) . $_compile_dir_sep . $_sha256;
         }
 
         return $_return;


Smarty/Smarty.class.php
```diff
@@ -1543,7 +1543,8 @@ function validateCompilePath( $compilePath ) {
         }
 
         $isValid = false;
-        if ( $fd = @fopen( $compilePath, 'wb') ) {
+        // Open file for writing but do not truncate.
+        if ( $fd = @fopen( $compilePath, 'c') ) {
             $isValid = true;
             @fclose( $fd );
             @unlink($compilePath);
```
totten: This isn't an issue with your patch, but a surprising observation about this section of code -- it looks like `validateCompilePath()` (*"do check can smarty create a file w/ given path" [sic]*) would effectively disable caching. If I'm reading correctly, every call to `Smarty::fetch()` calls `Smarty::_get_compile_path()`; as long as the resource uses `string:` notation, it will then call  `Smarty::validateCompilePath()` which then `@fopen`s and `@unlink`s the file... which means that cache file is always deleted before you get a chance to use it.

(Hrm... this whole scenario for caching `string:` seems overwrought...  as in... it's not sure what to name the cache-file, so it does `@fopen+@fclose+@unlink` to see if the filename produces a syntax error; and if there is an error, then it makes a uniqueish-but-not-reusable name with `time()+rand()`. Wouldn't it be simpler to have a more robust naming pattern like `$this->compile_dir . sha1($resource_name) . '.php'`?)

mfb:
That does sound more robust (I have no idea why it works this way :)  Oh and the other weird thing here is that it opens the file for writing.  And for a short period of time this compiled template is now on disk, completely empty, if it didn't already exist. That's not safe for concurrent processes either. For reference this was added in https://issues.civicrm.org/jira/browse/CRM-5890

Smarty/Smarty.class.php
mfb: That does sound more robust (I have no idea why it works this way :) Oh and the other weird thing here is that it opens the file for writing. And for a short period of time this compiled template is now on disk, completely empty, if it didn't already exist. That's not safe for concurrent processes either. For reference this was added in https://issues.civicrm.org/jira/browse/CRM-5890

mfb: It looks like even if you remove these lines, templates are still recompiled, because $smarty->compile_check is

tttp:
Hi,

What's the status on that one? we just go hit by the bus error (after upgrading to php7)

totten:
> ...not sure how to reliably test...

This.

I think if a couple folks like @mfb and @tttp have been using it in production for a while, then that's the best we can really do. Have you guys been using it?

tttp:
Yes we have been using it since a month ago in production, and sent 5M emails since the change.

@totten on our side: does work
@mfb did you try it?

mfb:
Yes we have to use this in production, otherwise we get the dreaded "bus error" and the process dies.  This resolves that issue, but occasionally (haven't checked stats, but it's rare) an individual mailing message is empty and CiviCRM therefore won't send it - this is a related concurrency issue due to template files being deleted and empty template files being created by multiple processes sharing the same templates.

mfb:
...This wasn't getting any traction, so I tried reworking it into something more robust: Using SHA-256 hash as the filename (rather than CRC-32 plus the string itself, which could be too long to be a valid filename, thus the weird creating and unlinking of compiled templates).

seamuslee001:
hmm I certainly haven't hit this before @totten This makes sense i think to me Tim, its also evident its in production in 2 separate sites and no errors found I think it should be merged on the basis of probabilities that its an improvement

mfb:
The current version of this PR might only be in use on one site, since I completely rewrote it a week ago.  Would be great to get a bit more testing. Maybe it's possible to run the test suite on a civicrm-packages PR?

seamuslee001:
@mfb i have just checked it out onto a local buildkit build of master and now going to try and run the unit tests on it locally. tho @totten maybe able to get it to run on Jenkins

seamuslee001:
@mfb i have just kicked off a Jenkins run based on this PR

seamuslee001:
@totten @mfb  unit tests passed on pr-158-1

mfb:
great!  I can't think of a possible regression since the code is mostly just simplified, but if there are any more volunteers willing to test that couldn't hurt..

mfb:
We've been using this in production so I think it's ready to be scheduled for next release.

seamuslee001:
@totten I would be a +1 on merging this, it has passed unit tests and appears to have been in production in one environment at least so i think it would be fairly stable

mfb:
I was looking at how Smarty generates keys for cached templates these days - looks like it uses sha1. However that was implemented some time ago, and given the SHAttered attack and general migration from sha1, sha256 seems reasonable to me.

xurizaemon:
Would SHA1 vs SHA256 make any difference? AFAICT we only use a small chunk of the hash anyway and they need to be predictable for use as a cache key. If we're seeing duplicate attempts to open the same file simultaneously, it's not the hash that's at fault MO.

(Not 100% but at a glance it seems like [Smarty is using CRC32](https://github.com/civicrm/civicrm-packages/blob/b57d6f400165ffc1b79af9df141b271365652864/Smarty/Smarty.class.php#L1765)?)

I've wondered about speed-testing `$smarty->setCompileDir('/dev/null');` against compile to disk, and I believe there is functionality or patches around to put `templates_c` in Redis etc. (Should also nicely address things like CRM-15632.)

mfb:
I used sha256 given that sha1 is generally deprecated (Twig and Drupal 8 for example use sha256)

seamuslee001:
Jenkins test this please

mfb:
We decided to disable CIVICRM_MAIL_SMARTY - doing so works around most instances of this bug (unless your site is really busy, I guess).  I wasn't planning to keep maintaining this PR since it's not critical for our instance, but this code is still "wrong" and should be fixed.. :)

mfb:
EFF folks are interested in re-enabling Smarty so that they can add more complex logic to email templates.  So there's a good chance I may resume work on this PR.

tttp:
Have you considered using a more modern template language? Eg mustache or
twig?

PR #152, opened 2016-06-06
WIP CRM-16897: Remove PEAR_Frontend_CLI class.

xurizaemon:
* [CRM-16897](https://issues.civicrm.org/jira/browse/CRM-16897)

PR #160, opened 2016-07-21
Remove PEAR CLI and various orphaned dependencies

totten:
We do not actively use PEAR CLI -- this code is effectively dead. Including it is more
of a liability (risk of misuse or misconfigurations leading to misuse) than a benefit.

For each of these libraries, I did a search across my local `master` build
(`civicrm-core`, `civicrm-drupal`, `civicrm-packages`).  However, it may be
prudent for others to double-check that.

Merge notes:
 * This must be merged *after* https://github.com/civicrm/civicrm-core/pull/8739
   to remove a dependency on a PEAR helper class.
 * I suggest merging this soon after 4.7.10-rc goes out.
 * PR #152 is a subset of this.
 * PR #150 may have some overlap (e.g. a handful of the classes are both cleaned in #150 and removed here).


totten:
Since this is a big set of changes, and since Jenkins doesn't run tests on `packages`, I've kicked off a manual test on on `test-ubu1204-5` using a manual build (`packages-160-1`) with output directed to `/tmp/packages-160-1.txt`

totten:
The above test run looks successful (`api_v3_AllTests`, `CRM_AllTests`, `Civi\\AllTests`, and `civicrm-upgrade-test` from 4.2.0 and 4.4.0).

There was one reported failure -- `api_v3_JobTest::testBatchMergeIDRange` -- but that particular test currently has an issue with inconsistent results.

totten:
So the test was fine. Just to be prudent, I did another round of searching to see if there were cases where these removed dependencies might be used.

I did come across some additional references to `Console_Getopt`. I suspect they may also be defunct, but I'll restore it for now.

PR #161, opened 2016-07-22
CRM-19116: Rewrite smarty_core_get_include_path to improve performance (4.7)

maevelevie:
See [CRM-19116: Smarty performance improvement](https://issues.civicrm.org/jira/browse/CRM-19116).

### DIFF
--- a/Smarty/internals/core.get_include_path.php
+++ b/Smarty/internals/core.get_include_path.php
@@ -7,36 +7,21 @@
 
 /**
  * Get path to file from include_path
+ * CRM-19116: Rewrite to improve performance
+ * stream_resolve_include_path exists in PHP >= 5.3.2.
  *
- * @param string $file_path
- * @param string $new_file_path
+ * @param string[] $params
+ * @param object $smarty
+ *
  * @return boolean
- * @staticvar array|null
  */
+function smarty_core_get_include_path(&$params, &$smarty) {
 
-//  $file_path, &$new_file_path
-
-function smarty_core_get_include_path(&$params, &$smarty)
-{
-    static $_path_array = null;
-
-    if(!isset($_path_array)) {
-        $_ini_include_path = ini_get('include_path');
-
-        if(strstr($_ini_include_path,';')) {
-            // windows pathnames
-            $_path_array = explode(';',$_ini_include_path);
-        } else {
-            $_path_array = explode(':',$_ini_include_path);
-        }
-    }
-    foreach ($_path_array as $_include_path) {
-        if (@is_readable($_include_path . DIRECTORY_SEPARATOR . $params['file_path'])) {
-               $params['new_file_path'] = $_include_path . DIRECTORY_SEPARATOR . $params['file_path'];
-            return true;
-        }
-    }
-    return false;
+  $newpath = stream_resolve_include_path($params['file_path']);
+  if (@is_readable($newpath)) {
+    $params['new_file_path'] = $newpath;
+    return TRUE;
+  }
 }
 
 /* vim: set expandtab: */


colemanw:
Curious to know if this function is covered by any unit test.

klevie:
Not specifically... But if this function wouldn't work, that would cause a lot of existing tests to fail  - especially most if not all WebTest. I'm not sure what a new unit test might add to that and how, but I'm open to suggestions.

totten:
The `civicrm-packages.git` repo does not automatically run tests, so I'm executing test suite manually on `test-ubu1204-5`. Will post back with results.

totten:
Unit tests pass.

Agree with the assessment that a lot of things would break if this weren't working.

FWIW, I also did some general 'click-around' testing and some manual function calls via CLI, e.g.:

```bash
cv ev 'require_once "Smarty/internals/core.get_include_path.php"; $p=array("file_path"=>"Civi.php"); smarty_core_get_include_path($p, CRM_Core_Smarty::singleton()); return $p;'

cv ev 'require_once "Smarty/internals/core.get_include_path.php"; $p=array("file_path"=>"CRM/DoesNotExist.php"); smarty_core_get_include_path($p, CRM_Core_Smarty::singleton()); return $p;'
```

Finally, I tried an experiment to see the affect on performance:

 * (Source code) https://gist.github.com/totten/576a73c3e7f6df890370ee743050fc1a
 * trials=`1`; runtime=`0.0470` (seconds); memory=`24848k` -- Original code
 * trials=`1`; runtime=`0.1054` (seconds); memory=`24841k` -- stream_resolve_include_path() without static cache
 * trials=`1`; runtime=`0.1542` (seconds); memory=`26457k` -- stream_resolve_include_path() plus static cache
 * trials=`100`; runtime=`3.5086` (seconds); memory=`24848k` -- Original
 * trials=`100`; runtime=`10.2580` (seconds); memory=`24841k` -- stream_resolve_include_path() without static cache
 * trials=`100`; runtime=`1.1289` (seconds); memory=`26457k` -- stream_resolve_include_path() plus static cache

This is actually a bit surprising -- in my benchmark, `stream_resolve_include_path()` takes twice as long! (For reference, I'm running on MAMP with `PHP 5.5.26 (cli) (built: Jul  6 2015 14:26:26)`.)

@klevie , any ideas on why this benchmark would have such different results?

totten:
1. In case there was a PHP difference, I tried PHP 5.3 instead. The older runtime was ~10% slower and consumed about 13mb more, but there was no change in ordering/relationships.

2. Some theories to provoke discussion:

 * Maybe a tool like Blackfire/Xdebug would show less time in `smarty_core_get_include_path()` but then show more time somewhere else (like `stream_resolve_include_path()` or a system library)?
 * Maybe my test script isn't really representative?

klevie:
Hmm, I think I can explain that... If I'm correct, in your test script the first call to `is_readable()` will succeed because the include path starts with the current directory. That might be quicker than a call to `stream_resolve_include_path()`.

On a real world installation, especially with a lot of extensions, both the PHP include path and Smarty resource paths are a lot longer. And I just discovered the Smarty `_parse_resource_name()` function includes another strange loop that makes it worse.
The number of calls to `is_readable()` then grows exponentionally: to load 30 template files, check 20 Smarty resource paths, and 20 PHP include directories, Smarty will call`is_readable()` for up to 12,000 times.

I can do some more tests / write a different test script / fix the other problematic Smarty function when I get back, in the first week of september.

totten:
Great, look forward to it.

Aside: Your profile mentions NL and CiviCoop. Perhaps we'll meet in October (during CiviCon or my visit to Erik).

klevie:
(Aside: I haven't heard what plans you've made with Erik yet :-) - but I'll definitely be at CiviCon and probably  also at the sprint. See you then! - And I'll post an update about this issue later as promised.)

seamuslee001:
Jenkins test this please

eileenmcnaughton:
We are finally getting off Smarty2 - I'm assuming they have figured out performance in later smarty versions

PR #163, opened 2016-07-30
CRM-17789 - Support PHP 7

totten:
This is a resubmission of @mollux's #150, but it's been rebased to resolve
some merge-conflicts.

---


PEAR.php
```diff
@@ -508,7 +508,7 @@ function delExpect($error_code)
      * @see PEAR::setErrorHandling
      * @since PHP 4.0.5
      */
-    function &raiseError($message = null,
+    static function raiseError($message = null,
```
totten: This ones concerns me a bit -- if you dig into this function, it's actually designed to work in static *or* non-static usages. And there are several non-static usages in, e.g., `Net_Socket` and `Mail_mimeDecode`. We need to be careful that we're not just trading one set of errors for another.

totten: It looks like upstream PEAR 1.10.0 has been patched with a clever work-around -- for any functions which were (by design) allowed to be static or non-static, these have been converted to magic-methods. I'll add the upgrade to this PR.

totten:
Now that we've merged https://github.com/civicrm/civicrm-core/pull/8790 and https://github.com/civicrm/civicrm-core/pull/8792, it *should* be possible to merge this one without any breaking-changes.

Running a manual test on `test-ubu1204-5` (`civibuild create pkgs-163-1 --patch https://github.com/civicrm/civicrm-packages/pull/163`).

totten:
Results from the manual run on test-ubu1204-5:

 * `civicrm-upgrade-test` (4.2.0, 4.4.0): Pass
 * `Civi\AllTests --exclude-group ornery`: Pass
 * `api_v3_AllTests --exclude-group ornery`: Pass
 * `CRM_AllTests --exclude-group ornery`: Mostly pass. There was inconsistent result in `CRM_Contact_BAO_ContactType_ContactTest::testCRM19133`. (It failed the first time I ran the suite. However, when I ran the suite a second time, it passed.)

seamuslee001:
@totten do you think we should merge this PR?

totten:
I'm doing some click-around testing of this PR and finding that report dashlets no longer render. Investigating...

totten:
Rebased and fixed issue on dashboard (showing OpenFlashChart).

DB/common.php
```diff
@@ -1873,7 +1873,7 @@ function dropSequence($seq_name)
      *
      * @see PEAR_Error
      */
-    function &raiseError($code = DB_ERROR, $mode = null, $options = null,
+    static function raiseError($code = DB_ERROR, $mode = null, $options = null,
```
totten: This smells a bit problematic to me
 * The parent class's `PEAR::raiseError()` is sorta polymorphic (static *or* nonstatic). In PEAR 1.10, this is done with magic functions, which is probably why it's not producing a syntax error.
 * There are only a couple classes which would inherit this function (`DB_mysql`, `DB_mysqli`, etc), and I'm not sure we haven't test-coverage that purposefully triggers a SQL error.

Maybe the right thing to do is override `protected static function _raiseError()`? (In PEAR 1.10, this is the real implementation shared by the static and non-static variants of `raiseError`.)

HTML/QuickForm.php
```diff
@@ -2008,7 +2008,7 @@ function isSubmitted()
      * @return bool     whether $value is an error
      * @static
      */
-    function isError($value)
+    public static function isError($value)
```
totten: Based on grepping, agree this one is really static.

HTML/QuickForm.php
```diff
@@ -2024,7 +2024,7 @@ function isError($value)
      * @return  string  error message
      * @static
      */
-    function errorMessage($value)
+    static function errorMessage($value)
```
totten: Based on grepping, agree this one is really static.

HTML/QuickForm/file.php
```diff
@@ -187,7 +187,7 @@ function onQuickFormEvent($event, $arg, &$caller)
                 break;
             case 'createElement':
                 $className = get_class($this);
-                $this->$className($arg[0], $arg[1], $arg[2]);
```
totten: That old code is somewhat scary construction.

Just to be safe, I grepped for other cases that might be like this (`$this->$foo`). None of the matches looked like obvious problems.

Mail/mimeDecode.php
```diff
@@ -621,7 +621,9 @@ function _quotedPrintableDecode($input)
         $input = preg_replace("/=\r?\n/", '', $input);
 
         // Replace encoded characters
-		$input = preg_replace('/=([a-f0-9]{2})/ie', "chr(hexdec('\\1'))", $input);
+        $input = preg_replace_callback('/=([a-f0-9]{2})/i', function ($matches) {
+            return chr(hexdec($matches[1]));
+        }, $input);
```
totten: I copied the old and new code to an isolated script and verified that they perform the same.

DB/common.php
totten: Poking at this further... 
 * It looks like `DB_common::raiseError($code, $mode,...)` has a rather different signature from the parent's `PEAR::raiseError($message,$code,$mode,...)`. E-v-i-l. Ee-vv-ii-ll. Eee-vvv-iii-lll.
 * PHP allows you to call `static` function using non-static notation (e.g. `$object->myStaticFunc()`), although `$this` does not propagate.
 * By adding `static` and obscuring `$this`, the patch actually makes `DB_common::raiseError()` behave differently -- it no longer reports the failed SQL query!
 * I can imagine why it was necessary to add `static` here when using PEAR 1.9 (i.e. there was a conflict in the function signatures); but in PEAR 1.10, upstream `PEAR::raiseError()` is actually a magic function which means it still runs even if the class-hierarchy suffers from conflicted signatures.

totten:
For the record, I also a did a test which basically involved (a) scanning for PHP files (b) loading all of them in one CLI script and (c) checking for errors in the result. There were some errors... but most of them were pre-existing or dealt with unused files (like PHPBeautifier's PEAR formatter).

seamuslee001:
@totten Tim do you think this PR is close now?

HTML/QuickForm/group.php
```diff
@@ -97,13 +97,15 @@ class HTML_QuickForm_group extends HTML_QuickForm_element
...
      *                                      the form $groupName[$elementName] or leave
      *                                      them as is.
+     * @param     mixed     $attributes     (optional)Either a typical HTML attribute string
+     *                                      or an associative array. Date format is passed along the attributes.
      * @since     1.0
      * @access    public
      * @return    void
      */
-    function __construct($elementName=null, $elementLabel=null, $elements=null, $separator=null, $appendName = true)
+    function __construct($elementName=null, $elementLabel=null, $elements=null, $separator=null, $appendName = true, $attributes = null)
```
totten: Spot-checked: I grepped for examples of `group`. Picked one from "Administer => CiviContribute => Premiums => ... => Edit => Premium Image" (a group of radio buttons). It still works in PHP 5.3.

HTML/QuickForm/hierselect.php
```diff
@@ -120,7 +120,7 @@ class HTML_QuickForm_hierselect extends HTML_QuickForm_group
      */
     function __construct($elementName=null, $elementLabel=null, $attributes=null, $separator=null)
     {
-        parent::__construct($elementName, $elementLabel, $attributes);
+        parent::__construct($elementName, $elementLabel, null, null, null, $attributes);
```
totten: Spot-checked: I grepped for examples of `hierselect`. Picked one from "Administer => Customize => ... => Add Custom Field" (data type)  and still works. It still works in PHP 5.3.

totten:
I was a bit nervous after finding that dashboard problem, so I continued reviewing this somewhat aggressively, e.g.
 * Continued with general click-around testing for this PR using PHP 5.3 (e.g. make a new "Contribution Page", "Price Set", and "Event"; view some reports with charts; add some premiums and custom fields; edit the custom fields on a contact). Haven't been able to find any regressions.
 * Reviewing each change on its own. This led to a few follow-up patches, although most were OK, and I struggle to find any more ways to test/review.
 * Ran the test-suite once more using a current build (latest `master` plus these patches). All the tests passed, except one (which I think is a red-herring).
  *  `CRM_Contribute_Form_ContributionTest::testSubmitCreditCardPayPal` failed.  However, I don't think the patchset is relevant. Specifically: That test fails on my local and on `test-ubu1204-5` (albeit with different errors). On both systems, the test-result is the same with or without the patch. So something is wrong with `ContributionTest`... but I don't think it's this patch...

So @seamuslee001 @mollux let's pull the trigger on this!

PR #165, opened 2016-08-31
Remove CreditCard Validation file as now provided by composer

seamuslee001:


seamuslee001:
Note that this is dependent on https://github.com/civicrm/civicrm-core/pull/8960

PR #166, opened 2016-09-09
CRM-16053 Upgrade ckeditor

seamuslee001:
* [CRM-16053: Upgrade CKEditor to latest version](https://issues.civicrm.org/jira/browse/CRM-16053)


seamuslee001:
@totten @colemanw @eileenmcnaughton Hey folks, would it be possible to get someone to test this, you will have to clear cache to make sure it you pick up the right js

colemanw:
@seamuslee001 4.7 already has the latest version of CKEditor. 4.6 is our LTS, which only gets critical/secuity fixes at this point afaik.

seamuslee001:
@colemanw yes but recently there was a security update for ckeditor and the version that 4.6 currently has is no longer supported in ckeditor i think.

colemanw:
Ah ok. Yes it will need some testing because this jumps up quite a few versions.

seamuslee001:
Just to note that i have put this onto an Australian Greens test server, have performed tests of re-using a mailing using inline style and I believe that we are satisfied that things are working fine.

PR #167, opened 2016-09-09
CRM-17789 switch to mysqli for dsn

seamuslee001:
ping @totten

---


totten:
This will allow `civicrm-core` to specify whether it wants to use literal DSN's (old behavior) or adaptive DSN's (new behavior). This makes it safe to merge the change to `packages` now while we're still reviewing the batch of `mysqli` changes in `core`.

PR #168, opened 2016-09-09
CRM-17789 Add in insertId function

seamuslee001:
ping @totten

---

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -4314,6 +4314,10 @@ class DB_DataObject extends DB_DataObject_Overload
     return '';
   }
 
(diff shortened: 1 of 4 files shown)


totten:
Thanks for opening a PR on this.

For quick manual testing, I used `cv scr` with this script, and it behaved correctly:

```php
<?php

$d = CRM_Core_DAO::executeQuery("insert into civicrm_log (entity_table,entity_id,data,modified_id,modified_date) values ('civicrm_foo', 123, NULL, NULL, now())");
var_dump(['id' => $d->getConnection()->lastInsertId()]);
```

PR #169, opened 2016-09-10
CRM-17789 Fix mysqli::free and update DB_DataObject free to what is in their rep…

seamuslee001:
…o and port mollux change to DB.php for free

### DIFF
--- a/DB/mysqli.php
+++ b/DB/mysqli.php
@@ -498,7 +498,11 @@ class DB_mysqli extends DB_common
      */
     function freeResult($result)
     {
-        return is_resource($result) ? mysqli_free_result($result) : false;
+       if (!$result instanceof mysqli_result) {
+            return false;
+        }
+        mysqli_free_result($result);
+        return true;
     }
 
     // }}}
--- a/DB.php
+++ b/DB.php
@@ -1403,12 +1403,9 @@ class DB_result
     function free()
     {
         $err = $this->dbh->freeResult($this->result);
-        if (DB::isError($err)) {
-            return $err;
-        }
         $this->result = false;
         $this->statement = false;
-        return true;
+        return $err;
     }
 
     // }}}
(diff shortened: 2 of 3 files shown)


seamuslee001:
@totten @mollux this is a resubmit of https://github.com/civicrm/civicrm-packages/pull/149 Note i have made slightly different changes to the free() in DB_DataObject. They match up to the source here http://svn.php.net/viewvc/pear/packages/DB_DataObject/trunk/DB/DataObject.php?revision=339312&view=markup

totten:
Did a test run on `test-ubu1204-5` with `pkgs-169-1` (`CRM`, `api_v3`, `Civi`, and `civicrm-upgrade-test` w/4.2.0 or 4.4.0). All those tests passed, except `CRM_Contribute_Form_ContributionTest::testSubmitCreditCardPayPal` which has been ornery lately.

seamuslee001:
@totten so we can merge?

DB/mysqli.php
```diff
@@ -498,7 +498,11 @@ function fetchInto($result, &$arr, $fetchmode, $rownum = null)
      */
     function freeResult($result)
     {
-        return is_resource($result) ? mysqli_free_result($result) : false;
```
eileenmcnaughton: I have had to revert this for wmf to get it to pass their tests . I'm not sure if the problem is on the servers - or only the test env (running hhvm) - https://integration.wikimedia.org/ci/job/wikimedia-fundraising-civicrm/3756/console

seamuslee001: Feels strange this changes matches the core project tho

eileenmcnaughton: I didn't actually find the core file - the link above doesn't have that line in it

seamuslee001: see https://github.com/pear/DB/blob/trunk/DB/mysqli.php#L498

PR #171, opened 2016-09-15
CRM-17789 4.6 packages support for php7

seamuslee001:
- [CRM-17789: Support PHP 7](https://issues.civicrm.org/jira/browse/CRM-17789)


jackrabbithanna:
We're considering this patch for inclusion into the LTS at some point this year (no rush).  @guanhuan let me know if you have any objections.  There would need to be a lot of testing.  We have one client that wants it, and more may come...so I'm thinking about making it happen for them (before merging) and letting it run in a production site for while.  

@seamuslee001 if you catch wind of anything related to this that breaks in 4.7 and needs fixing, let us know so we can make this a possibility for 4.6

seamuslee001:
@jackrabbithanna Mark i did a manual test run of this PR on the test-ubu1204-5 and it passed.

seamuslee001:
Jenkins test this please

seamuslee001:
I'm going to close this out as 4.6 is no longer supported

PR #172, opened 2016-09-20
CRM-17789 - Fix deprecate php4-style constructors for Log.php.

jitendrapurohit:
>Deprecated function: Methods with the same name as their class will not be constructors in a future version of PHP; Log has a deprecated constructor in require_once() (line 38 of /home/web/civi_master/civicrm/CRM/Core/Config.php).

Refer - https://github.com/pear/Log/blob/master/Log.php#L114

---

### DIFF
--- a/Log.php
+++ b/Log.php
@@ -101,6 +101,10 @@ class Log
                             '%{class}'      => '%8$s',
                             '%\{'           => '%%{');
 
+    public function __construct()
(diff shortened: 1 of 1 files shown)


totten:
Oooh, yeah. I'd seen that warning too and wasn't sure how to solve it. Nice.

PR #173, opened 2016-10-24
Disable peer verification for SMTP by default

guanhuan:
Problem came up when we were trying to test some new mailing services with CiviCRM 4.7, we kept getting "connection refused" errors and other similar messages with different protocols and ports entered in outbound mail settings.

The environment we have: Ubuntu 14.04, php 5.6, Drupal 7.3.x, CiviCRM 4.7.x.

The problem turns out to be related to php 5.6. Since php has enabled peer verification by default from 5.6.x, all the SMTP connection via SSL and TLS will fail in CiviCRM.

There is a temporary fix for this. Change in packages/Net/SMTP.php

line $this->_socket_options = $socket_options;

to $this->_socket_options = array('ssl' => array('verify_peer_name' => false));

The above solves the problem with SMTP failure however we should either be updating CiviCRM or turn off the peer verification elsewhere in php setting. Otherwise, the document here probably should updated to point out 5.6 compatibility issue.

My answer for ref: http://civicrm.stackexchange.com/questions/10964/smtp-connection-failing-problem/10966#10966

### DIFF
--- a/Net/SMTP.php
+++ b/Net/SMTP.php
@@ -184,7 +184,7 @@ class Net_SMTP
         $this->pipelining = $pipelining;
 
         $this->_socket = new Net_Socket();
-        $this->_socket_options = $socket_options;
+        $this->_socket_options = array('ssl' => array('verify_peer_name' => false));
         $this->_timeout = $timeout;
 
         /* Include the Auth_SASL package.  If the package is available, we 


seamuslee001:
This feels wrong to me, I think you should be ensuring that your mail server etc has correct ssl certs

guanhuan:
Maybe a better idea is to allow this option to be overridden somewhere in civicrm.settings.php but leave it on by default?

mlutfy:
Does your SMTP server have a valid TLS configuration?

I ran into this issue a long time ago, but using: `ssl://smtp.example.org` and port 465, it usually works OK.

guanhuan:
@mlutfy It is not an issue that bothers me any more but if I remember correctly we have a valid SSL configured for the site and CiviCRM outbound is configured with authentication and port 465/25. It stopped working upon upgrading to Civi 4.7 and php 5.6 (with same settings). I have updated my answer in stackexchange to reflect that I agree by turning off peer verification is not a good solution but I think by having an option to override might provide convenience to people who need it (for test environment etc.). Just a suggestion if anyone who is interested to work on this.

guanhuan:
@mlutfy, I believe the issue came up on the site that has SSL related configurations already set up including `ssl://` since that is our regular way to setting up outbound emails.

PR #178, opened 2017-04-11
CRM-20178 & CRM-20414 apply db level fixes to update (only applied to insert)

eileenmcnaughton:
* [CRM-20178: Writes to timestamp fields are silently ignored by $dao-\>save\(\)](https://issues.civicrm.org/jira/browse/CRM-20178)
 * [CRM-20414: Writes to date & timestamp fail on insert but not update - this is a follow on from CRM-14986](https://issues.civicrm.org/jira/browse/CRM-20414)

---

 * [CRM-14986: fatal error when trying to cancel a contribution recur payment](https://issues.civicrm.org/jira/browse/CRM-14986)

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -968,11 +968,6 @@ class DB_DataObject extends DB_DataObject_Overload
             if (!isset($this->$k)) {
                 continue;
             }
-            // dont insert data into mysql timestamps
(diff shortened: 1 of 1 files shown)


monishdeb:
Reviewed the patch and the related changes in core https://github.com/civicrm/civicrm-core/pull/10137. Working fine, merging it now

PR #181, opened 2017-04-29
typo in error message

mhurwicz:


xurizaemon:
Our copy of `Net/SMTP.php` is dated, you might want to file this fix against the upstream repo as well?

https://github.com/pear/Net_SMTP/blob/master/Net/SMTP.php

eileenmcnaughton:
good spotting!

PR #182, opened 2017-05-10
(WIP) CRM-18408 - Net_SMTP - Use CA_Config for SSL options

totten:
Untested

---


totten:
This was flagged WIP because it was based on a theory about the cause of CRM-20561. Testing showed that the actual problem was different, and this isn't really needed.

PR #183, opened 2017-05-11
(WIP) CRM-18408, update for TLS issues seen in PHP5.6

xurizaemon:
PHP5.6 introduces peer validation by default - this is a good thing.

But it makes some TLS negotiation issues visible too. This ports changes from Net_SMTP-1.7.1 to our ancient patched 2011 version.

We could also update Net_SMTP properly, but our version is modified so that wants review.

This works for the specific site I was able to test (Elasticemail).

This commit accidentally includes 1f656ec which should remain but be separately committed.

---

### DIFF
--- a/Net/SMTP.php
+++ b/Net/SMTP.php
@@ -183,6 +183,14 @@ class Net_SMTP
         }
         $this->pipelining = $pipelining;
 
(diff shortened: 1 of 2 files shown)


xurizaemon:
Closing this PR in favour of #184

PR #184, opened 2017-05-11
Update for PHP5.6.x changes with regard to TLS.

xurizaemon:
In PHP5.6.7 there are changes to the constants relating to TLS stream socket options.

This basically pear/Net_SMTP@90de8451.

CRM-18408

---


totten:
Addressed by https://github.com/civicrm/civicrm-core/pull/10384

PR #188, opened 2017-06-12
CRM-20243 Remove duplicated DataTables plugin its provided by Bower

seamuslee001:
* [CRM-20243](https://issues.civicrm.org/jira/browse/CRM-20243)


seamuslee001:
ping @totten

seamuslee001:
Dang nice pick up Makes me wonder if we should be updating that to be the relevant version provided by bower. Tho what i think we can do is just revert the removal of those 2 files specifically

seamuslee001:
@totten i have not reverted the removal of the media/images folder which i think will solve the issue you found

seamuslee001:
@totten is this ok now? btw coleman has merged my PR to update the css file but still does mean the issue with those who have copied is still an issue

totten:
Grepped `universe`. Can't find any more lingering references, so should be good.

PR #190, opened 2017-06-15
CRM-20561 Remove Net_SMTP and Net_Socket as provided by composer now

seamuslee001:
* [CRM-20561: Load Net_SMTP, Auth_SASL, Net_Socket via Composer](https://issues.civicrm.org/jira/browse/CRM-20561)


seamuslee001:
ping @totten

PR #192, opened 2017-06-28
CRM-10414: return-path overrides from prematurely

monishdeb:
* [CRM-10414: return-path overrides from prematurely](https://issues.civicrm.org/jira/browse/CRM-10414)

### DIFF
--- a/Mail/smtp.php
+++ b/Mail/smtp.php
@@ -36,7 +36,7 @@
  *
  * @category    HTTP
  * @package     HTTP_Request
- * @author      Jon Parise <jon@php.net> 
+ * @author      Jon Parise <jon@php.net>
  * @author      Chuck Hagenbuch <chuck@horde.org>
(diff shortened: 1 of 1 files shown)


lcdservices:
looks good.

seamuslee001:
Jenkins test this please

eileenmcnaughton:
@mattwire @seamuslee001 feels like the sort of thing you might have an interest in / be prepared to review

mattwire:
Change probably makes sense, though I haven't spent enough time to completely understand the detail.  One comment though, we're using quite an old version of that library and the latest upstream version has not made this change: https://github.com/pear/Mail/blob/master/Mail/smtp.php#L292 @monishdeb I'd feel happier if an upstream dev was able to verify the fix?

eileenmcnaughton:
This file doesn't exist any more! I guess we moved it out of packages. On the bright side that probably means there is a vendor package for it that might be being updated....

PR #194, opened 2017-08-12
CRM-21059 Upgrade Smarty to 2.6.30

seamuslee001:
* [CRM-21059](https://issues.civicrm.org/jira/browse/CRM-21059)

### DIFF
--- a/Smarty/Config_File.class.php
+++ b/Smarty/Config_File.class.php
@@ -29,7 +29,7 @@
  * @package Smarty
  */
 
-/* $Id: Config_File.class.php 3149 2009-05-23 20:59:25Z monte.ohrt $ */
+/* $Id$ */
 
 /**
  * Config file reading class
@@ -73,7 +73,7 @@ class Config_File {
      *
      * @param string $config_path (optional) path to the config files
      */
-    function __construct($config_path = NULL)
+    public function __construct($config_path = NULL)
     {
         if (isset($config_path))
             $this->set_path($config_path);
--- a/Smarty/Config_File.class.php
+++ b/Smarty/Config_File.class.php
@@ -19,7 +19,7 @@
  *
  * For questions, help, comments, discussion, etc., please join the
  * Smarty mailing list. Send a blank e-mail to
- * smarty-discussion-subscribe@googlegroups.com 
+ * smarty-discussion-subscribe@googlegroups.com
  *
  * @link http://www.smarty.net/
  * @version 2.6.25-dev
@@ -301,7 +301,7 @@ class Config_File {
                         $vars = array();
                         continue;
                     }
-                } else {                    
+                } else {
                     $section_name = $match[1];
                 }
                 if (!isset($config_data['sections'][$section_name]))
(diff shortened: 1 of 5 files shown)


totten:
jenkins, test this please

seamuslee001:
jenkins, test this please

seamuslee001:
@totten i have cleaned up a bunch of the whitespace issues now and i think this should be good to merge

seamuslee001:
jenkins, test this please

seanmadsen:
I have reviewed this PR and I think it's good to merge. :white_check_mark: 

Here are some more detailed notes.

## I looked at the change descriptions

Since we are upgrading from Smarty 2.6.27 to 2.6.30, I looked for release notes for 2.6.28, 2.6.29, and 2.6.30. The best thing I found was this [change log](https://github.com/smarty-php/smarty/blob/Smarty2/ChangeLog) and I looked at the changes between 2013-09-30 and now which are listed below. *(Note: I also looked for a better change log [here](https://github.com/smarty-php/smarty/blob/master/change_log.txt) and [here](https://github.com/smarty-php/smarty/releases/tag/v2.6.28) but couldn't find one.)* Changes are as follows:

1. Fixed made Smarty_Compiler.class.php compatible with PHP 5.5
2. Fixed old vulnerability bug https://bugs.gentoo.org/show_bug.cgi?id=356615
3. PHP7 raises E_DEPRECATED use __construct for compatibility
4. fixed plugin filepath cache must not be static, because of possible problem when using multiple Smarty instances with diffrent plugins_dir settings https://github.com/smarty-php/smarty/issues/146
5. {math} shell injection vulnerability patch provided by Tim Weber
6. {math} fix parameter checking order to avoid misleading message
7. {math} replace wrong versiom


For the most part, all of these changes seem very unlikely to introduce any issues for us.

I paid particular attention to `4.` and `5.`...

### `plugins_dir` change

I read about this change and looked in our codebase for places where we set the `plugins_dir`. I'm pretty sure this change will not impact us.

### `{math}` change

Backticks and dollar signs are now disallowed in math equations. I used the following command to see what math equations our code (universe) is using:

```
$ grep -Piroh '\{math ((?!\})(.|\n))+\}' ~/buildkit/build/universe | grep -Pio '(?<=equation=")((?!")(.|\n))+'
x / y
x / y
x / y
x / y
(x/y) * 100
(x/y) * 100
(x - y)+2
x/y
150-(100*x)
100*x
x + y
x - y
```

No backticks. No dollar signs. I think we're good here.

Just to be sure. I looked deeper into the two locations in core that use `{math}` ([here](https://github.com/civicrm/civicrm-core/blob/2a73d3b05ba42e8119b7123ba93f6a7f5a2dbac4/templates/CRM/Pledge/Form/Pledge.tpl#L38) and [here](https://github.com/civicrm/civicrm-core/blob/2a73d3b05ba42e8119b7123ba93f6a7f5a2dbac4/templates/CRM/Pledge/Form/PledgeView.tpl#L26)). I clicked through some pledge forms that should be hitting those formulas. Everything went as expected.

## I looked at the diff

Nothing stood out

eileenmcnaughton:
I'm happy with this analysis/review - we should give an extra plug for testing during Rc period

eileenmcnaughton:
Also - I have a feeling I might have used {math} in workflow templates before - - we could add a check that gives a message if the string {math} is found in the message_template table?

PR #196, opened 2017-11-21
CRM-21458: Replace addcslashes call with json_encode creating options

agileware-fj:
json_encode should ensure that *all* escapes required for JS string values
are present.

* CRM-21458: (CIVICRM-742) HTML_QuickForm_hierselect doesn't handle JS escaping properly
  https://issues.civicrm.org/jira/browse/CRM-21458

### DIFF
--- a/HTML/QuickForm/hierselect.php
+++ b/HTML/QuickForm/hierselect.php
@@ -362,7 +362,7 @@ class HTML_QuickForm_hierselect extends HTML_QuickForm_group
             if ($js != '') {
                 $js .= ",\n";
             }
-            $js .= '"'.$optValue.'":"'.json_encode($options).'"';
+            $js .= '"'.$optValue.'":'.json_encode($options);
         }
     }
 
--- a/HTML/QuickForm/hierselect.php
+++ b/HTML/QuickForm/hierselect.php
@@ -362,7 +362,7 @@ class HTML_QuickForm_hierselect extends HTML_QuickForm_group
             if ($js != '') {
                 $js .= ",\n";
             }
-            $js .= '"'.$optValue.'":"'.addcslashes($options,'"').'"';
+            $js .= '"'.$optValue.'":"'.json_encode($options).'"';
         }
     }
 


eileenmcnaughton:
test this please

agileware-fj:
From http://php.net/manual/en/function.json-encode.php:

>  Like the reference JSON encoder, json_encode() will generate JSON that is a simple value (that is, neither an object nor an array) if given a string, integer, float or boolean as an input value. While most decoders will accept these values as valid JSON, some may not, as the specification is ambiguous on this point. 

Because the output of this function is actually generating an inline script to be loaded on the page, we don't need to be worried about JSON parsing compatibility, while enjoying the benefit of a string that is encoded to work correctly with Javascript.

totten:
(*CiviCRM Review Template WORD-1.0*)



* ([`r-jira`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-jira)) __Pass__
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) __Pass__
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __Pass__: I was a little concerned because `civicrm-packages` traditionally had a weird protocol wherein one was expected to copy the *old* file to `_ORIGINAL_` (to indicate that it's been forked). Fortunately, that's already been done. 👍 
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __Pass__
* ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __Pass__
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __Pass__: Tested the steps described in CRM-21458. Couldn't think of anything else to try.
* ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __Pass__: Generally preserves behavior of UI.
* ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __Pass__: 
   * This only affects the UI.
   * I suppose it's possible that someone might have put funny codes in their titles (in their data) to work-around this bug. However, that seems unlikely given that `addcslashes()` was already active (which would have covered several edge-cases and also would interfere with manual workarounds).

PR #198, opened 2018-01-10
CRM-17352 - Use Backbone.noConflict

colemanw:
A dependency for https://github.com/civicrm/civicrm-core/pull/11504

---

 * [CRM-17352: Backbone no-conflict](https://issues.civicrm.org/jira/browse/CRM-17352)


backbone/backbone.collectionsubset.js
```diff
@@ -230,4 +230,4 @@
     module.exports = Backbone.CollectionSubset;
   }
 
-}(_));
+}(_, Backbone));
```
dsnopek: Shouldn't this be `CRM.BB` rather than just `Backbone` to ensure that CiviCRM's version will be used? Or, are you depending on this running before the CMS will have added a different version of Backbone to the page?

colemanw: Right, this script goes in the header, so it should have access to `Backbone`. This works in D7 and WP so we need to track down what's going on with D8.

PR #200, opened 2018-04-05
(accessibility/misc-core-fixes/issues/3) Add aria-label attribute to Quickform elements

monishdeb:
Overview
This PR is about adding accessible labels by adding ```aria-label``` attribute to all kind of input fields.
Ticket: https://lab.civicrm.org/accessibility/misc-core-fixes/issues/3

Before
Public form - online contribution page
screen shot 2018-04-05 at 12 43 48 pm

After
Public form - online contribution page
screen shot 2018-04-05 at 12 42 38 pm

### DIFF
--- a/HTML/QuickForm/element.php
+++ b/HTML/QuickForm/element.php
@@ -3,7 +3,7 @@
 
 /**
  * Base class for form elements
- * 
+ *
  * PHP versions 4 and 5
  *
  * LICENSE: This source file is subject to version 3.01 of the PHP license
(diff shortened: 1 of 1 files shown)


monishdeb:
@colemanw I am using WAVE chrome plugin to evaluate which element is missing the accessible label. The ```for``` attribute for ```<label>``` tag already serve the purpose but in some cases there is no corresponding label for an input field. And thus this implicitly adds the ```aria-label``` to indicate the label. 

 Thanks for your suggestion, I have updated the PR.

PR #201, opened 2018-04-11
Update DB.php

Jazz-Man:
compatibility with php7 has been updated

### DIFF
--- a/DB.php
+++ b/DB.php
@@ -451,7 +451,7 @@ class DB
      *
      * @see DB_common::setOption()
      */
-    function &factory($type, $options = false)
+    public function &factory($type, $options = false)
     {
         if (!is_array($options)) {
             $options = array('persistent' => $options);
@@ -478,7 +478,7 @@ class DB
 
         foreach ($options as $option => $value) {
             $test = $obj->setOption($option, $value);
-            if (DB::isError($test)) {
+            if (self::isError($test)) {
                 return $test;
             }
         }
@@ -522,9 +522,9 @@ class DB
      *
      * @uses DB::parseDSN(), DB_common::setOption(), PEAR::isError()
      */
-    static function &connect($dsn, $options = array())
+    public static function &connect($dsn, $options = array())
     {
-        $dsninfo = DB::parseDSN($dsn);
+        $dsninfo = self::parseDSN($dsn);
         $type = $dsninfo['phptype'];
 
         if (!is_array($options)) {
@@ -556,13 +556,13 @@ class DB
 
         foreach ($options as $option => $value) {
             $test = $obj->setOption($option, $value);
-            if (DB::isError($test)) {
+            if (self::isError($test)) {
                 return $test;
             }
         }
 
         $err = $obj->connect($dsninfo, $obj->getOption('persistent'));
-        if (DB::isError($err)) {
+        if (self::isError($err)) {
             if (is_array($dsn)) {
                 $err->addUserInfo(DB::getDSNString($dsn, true));
             } else {
@@ -582,7 +582,7 @@ class DB
      *
      * @return string  the DB API version number
      */
-    function apiVersion()
+    public function apiVersion()
     {
         return '1.7.13';
     }
@@ -597,7 +597,7 @@ class DB
      *
      * @return bool  whether $value is DB_Error object
      */
-    static function isError($value)
+    public static function isError($value)
     {
         return is_a($value, 'DB_Error');
     }
@@ -612,7 +612,7 @@ class DB
      *
      * @return bool  whether $value is a DB_<driver> object
      */
-    function isConnection($value)
+    public function isConnection($value)
     {
         return (is_object($value) &&
                 is_subclass_of($value, 'db_common') &&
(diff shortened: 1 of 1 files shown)


mlutfy:
jenkins, test this please

DB.php
```diff
@@ -790,7 +790,7 @@ static function parseDSN($dsn)
             $parsed['dbsyntax'] = $str;
         }
 
-        if (!count($dsn)) {
+        if (($dsn instanceof \Countable /*|| is_array($dsn)*/) && !count($dsn)) {
```
totten: Most of the changes look pretty safe and make the code feel a bit more modern/normal. 👍 

Could you describe a little more about the reasoning behind this one?

Skimming some of the code above, it seems like array-based DSNs would have already returned (line 771), leaving `$dsn` as a string or null (line 777 or 780). I sorta wonder if they actually meant to use `strlen()` instead of `count()`... or if the general flow changed since the `count()` was put in? I'm just ranting here, but basically I'd like to understand this line better.

Jazz-Man: Frankly, this code was not very clear to me. But after testing on php72 I was constantly getting an error "Warning: count(): Parameter must be an array or an object that implements Countable in %s on line %d".
After that I decided to debug the code. On php71 and below, the code on line 794 never returned an array with $parsed. During testing on php72 on the same line the array returned,
in which the value was assigned only for "phptype" and "dbsyntax", and the remaining values are respectively "false". I also tried to enter the array with the data to connect to the database in the constants CIVICRM_UF_DSN and CIVICRM_DSN. But I had a problem, because here https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Config/Runtime.php#L178 it is necessary that the constant contains a string. Guided by the documentation https://pear.php.net/manual/en/package.database.db.intro-dsn.php I'm now trying to write a function based on parse_url
```php
$parsed = [
  'phptype'  => false,
  'dbsyntax' => false,
  'username' => false,
  'password' => false,
  'protocol' => 'tcp',
  'hostspec' => false,
  'port'     => false,
  'socket'   => false,
  'database' => false,
];


$dsn = $test_dns = "mysqli(mysqli)://{$db[$env]['username']}:{$db[$env]['password']}@{$db[$env]['host']}:5555/{$db[$env]['database']}?new_link=true";

$test_dns = rawurldecode($test_dns);
$test_dns = parse_url($test_dns);

$_regex_db_type = '/(.+?)\((.*?)\)/';

if (preg_match($_regex_db_type, $test_dns['path'], $arr)) {
    $parsed['phptype']  = $arr[1];
    $parsed['dbsyntax'] = !$arr[2] ? $arr[1] : $arr[2];

    $_db_dns = preg_replace($_regex_db_type, $parsed['phptype'], $dsn);
    $test_dns = parse_url($_db_dns);

} else {
    $parsed['phptype'] = $parsed['dbsyntax'] = $test_dns['scheme'];
}

if (!empty($test_dns['port'])){
    $parsed['port'] = $test_dns['port'];
}

if (!empty($test_dns['user'])){
    $parsed['username'] = $test_dns['user'];
}

// Find protocol and hostspec

$_regex_proto_and_host = '/@(.*?)\((.*?)\)/';

if (preg_match($_regex_proto_and_host, $dsn, $match)) {

    list($_dsn,$protocol ,$socket) = $match;

    $parsed['protocol'] = $protocol;

    if ($parsed['protocol'] === 'unix'){
        $parsed['socket'] = $socket;
    }
}

// $dsn => protocol+hostspec/database (old format)
if (strpos($test_dns['host'], '+') !== false) {

    list($protocol,$host) = explode('+', $test_dns['host'], 2);

    $parsed['protocol'] = $protocol;

    switch ($parsed['protocol']){
        case 'tcp':
            $parsed['hostspec'] = $host;
            break;
        case 'unix':
            $parsed['socket'] = $host;
            break;

    }

}else{
    $parsed['hostspec'] = $test_dns['host'];
}

if (!empty($test_dns['pass'])){
    $parsed['password'] = $test_dns['pass'];
}


dump([
  $test_dns,
  $dsn,
  $parsed,
    parse_url('phptype:///database?option=value&anotheroption=anothervalue')
]);

```

But I have doubts about which format of the string in civicrm was used earlier to connect to the database and therefore in some parts of the code I use old regular expressions.
My function is far from over, but I think that we will coordinate everything with you here:)

seamuslee001:
Jenkins re test this please

DB.php
seamuslee001: @Jazz-Man @totten how about we change count to be strlen instead, Reason being is that at L771 i think it is we return `$dsn` if its an array after merging in some defaults so the only other thinking would be is that its now a string hence the error

nganivet:
@totten @seamus This PR can probably be closed as DB.php already has similar patches in place for PHP 7.x compatibility.

totten:
Yup, agree we can close this.  Cheers @nganivet

Nowadays, the tests are regularly running+passing on a few increments of PHP 7.x

There are some bits (e.g. adding `public` flags) that are nice and which aren't upstream, but they also don't have functional effect. I think if upstream cleans those bits up, then it'd make sense to follow suit. However, we don't need to hold open.

PR #203, opened 2018-04-26
Allow KCFinder to send back a json encoded response instead of string…

seamuslee001:
… javascript

ping @colemanw

### DIFF
--- a/kcfinder/core/class/uploader.php
+++ b/kcfinder/core/class/uploader.php
@@ -87,6 +87,10 @@ class uploader {
   * @var string */
     protected $cms = "";
 
+/** Ouput format.
+  * @var string */
+    protected $outputFormat = 'js';
+
 /** Magic method which allows read-only access to protected or private class properties
   * @param string $property
   * @return mixed */
@@ -247,6 +251,11 @@ class uploader {
                 : path::url2fullPath($this->config['uploadURL']);
             $this->typeDir = "{$this->config['uploadDir']}/{$this->type}";
             $this->typeURL = "{$this->config['uploadURL']}/{$this->type}";
+        }
+
+        // Output Format
+        if (isset($_GET['format'])) {
+          $this->outputFormat = $_GET['format'];
         }
 
         // HOST APPLICATIONS INIT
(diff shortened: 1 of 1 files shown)


kcfinder/core/class/uploader.php
```diff
@@ -759,11 +768,19 @@ protected function callBack($url, $message="") {
                 $js = $this->$method($url, $message);
         }
 
-        if (!isset($js))
+        if ($this->outputFormat == 'json') {
+          $json = $this->callBack_json($url, $message);
+          echo json_encode($json);
```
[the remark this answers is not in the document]
seamuslee001: @colemanw what would you recommend for content type?

kcfinder/core/class/uploader.php
```diff
@@ -759,11 +768,20 @@ protected function callBack($url, $message="") {
...
+          $js = $this->callBack_ckeditor($url, $message);
+          if (!isset($js)) {
             $js = $this->callBack_default($url, $message);
 
-        header("Content-Type: text/html; charset={$this->charset}");
-        echo "<html><body>$js</body></html>";
+            header("Content-Type: text/html; charset={$this->charset}");
+            echo "<html><body>$js</body></html>";
+          }
+        }
```
colemanw: There's something odd going on here. The deleted line on 762 has no opening bracket, but the line you aded on 778 does. I think that may be causing the odd behavior.

kcfinder/core/class/uploader.php
```diff
@@ -759,11 +768,19 @@ protected function callBack($url, $message="") {
...
+          echo json_encode($json);
+        }
+        else {
+          if (!isset($js)) {
             $js = $this->callBack_default($url, $message);
 
-        header("Content-Type: text/html; charset={$this->charset}");
-        echo "<html><body>$js</body></html>";
+            header("Content-Type: text/html; charset={$this->charset}");
+            echo "<html><body>$js</body></html>";
```
colemanw: The above 2 lines should not be inside that conditional. The original version of this file only has 1 line "inside" the conditional (which is what happens when an if statement doesn't have brackets.

seamuslee001:
thanks @colemanw should be fixed now me thinks

colemanw:
This looks good now and is working correctly according to my testing.

PR #208, opened 2018-05-06
Upgrade Smarty to 2.6.31 to solve issues on PHP7.2

seamuslee001:


Smarty/Smarty_Compiler.class.php
```diff
@@ -556,7 +551,7 @@ function _compile_tag($template_tag)
 
             case 'php':
                 /* handle folded tags replaced by {php} */
-                list(, $block) = each($this->_folded_blocks);
+                $block = array_shift($this->_folded_blocks);
```
totten: Most of the changes look like clear equivalents. The loop through `$this->_folded_blocks` is the only one which feels like it *could potentially* have side-effects (depending on how `$this->_folded_blocks` is used elsewhere), but I'd defer to upstream if they like it.

totten:
(*CiviCRM Review Template WORD-1.1*)



* ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) __Pass__: The purpose and substance of the upgrade seem narrow, so I don't think it needs much more detailed explanation.
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) __Pass__
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) __Pass__
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) __Pass__: Not needed
* ([`r-maint`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-maint)) __Pass__: same as before
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) __Pass__: Spot-checked some pages which use Smarty in different ways (e.g. plain pages and AJAX snippets).
* ([`r-user`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-user)) __Pass__: No UX implication
* ([`r-tech`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-tech)) __Pass__: No change in contract

Thanks, @seamuslee001 !

PR #209, opened 2018-05-09
Strip trailing whitespace in quickform package

eileenmcnaughton:
We have a situation where we have a proposed change (from @monishdeb ) that is hard to read because of the trailing spaces being stripped - this is hard to avoid for anyone using an IDE or decent editor since trailing whitespaces are against our standard.

OTOH the upstream QF package is basically dead so we don't gain from keeping in sync with it.

It is easier to review this if you add w=1 to the url - ie. https://github.com/civicrm/civicrm-packages/compare/master...eileenmcnaughton:white?expand=1&w=1

This is not a full whitespace tidy up - in fact line endings are still set to windows :-(
But, by stripping the trailing ones we make legible PRs much more possible

### DIFF
--- a/HTML/Common.php
+++ b/HTML/Common.php
@@ -11,7 +11,7 @@
  * http://www.php.net/license/3_01.txt If you did not receive a copy of
  * the PHP License and are unable to obtain it through the web, please
  * send a note to license@php.net so we can mail you a copy immediately.
- * 
+ *
  * @category    HTML
  * @package     HTML_Common
  * @author      Adam Daniel <adaniel1@eesus.jnj.com>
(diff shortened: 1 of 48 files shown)


seamuslee001:
I believe its ok to merge as is, I think trailing whistpace is just something we should do with upgrades giving this merge on pass as all changes are whitespace only

eileenmcnaughton:
test bot killed but was running ok

monishdeb:
Ahh right, it's my IDE atom which automatically strips whitespace whenever I open a file and thus make the rightful change which I realised later on ```git diff```. Thanks @eileenmcnaughton @seamuslee001 now I don't have worry about whitespaces in QF and my patch would be much cleaner :)

eileenmcnaughton:
@monishdeb - you are not alone with a whitespace stripping IDE - mine wanted to convert line endings too but I kept that out

PR #205, opened 2018-05-05
Upgrade Mime_mail to fix issues with PHP7.2

seamuslee001:

PR #210, opened 2018-05-15
Declare Subparts as array in mimePart to support PHP7.2

seamuslee001:
@eileenmcnaughton @monishdeb this pulls out the minor commit from https://github.com/civicrm/civicrm-packages/pull/205 as the full upgrade appears to be causing significant test failures. This just defines subparts as an array as needed  to pass php7.2 count() issues

### DIFF
--- a/Mail/mimePart.php
+++ b/Mail/mimePart.php
@@ -89,7 +89,7 @@ class Mail_mimePart
     * @var array
     * @access private
     */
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
I took a look at how the property is used & this seems safe & correct

seamuslee001:
merging as per tag

PR #211, opened 2018-05-17
dev/core#117 Replace useage of deprecated function each with foreach …

seamuslee001:
…in Mail/mime Mail/mimePart and Mail/mimeDecode

This is again another small pull out rather than trying to do the full upgrade of these packages as the full upgrade is having some issues This just removes the each() function in favor of foreach()


monishdeb:
Changes are straightforward. Reviewed the underlying code and tested mail workflow, works fine.

PR #212, opened 2018-05-23
Standardise line endings away from windows line endings to linux

eileenmcnaughton:
Replace windows line endings with linux ones in Quickform package.

This is our standard & having windows line endings makes the code hard to create clean diffs on.

QF is not maintaining upstream so don't need to worry about keeping in sync

The best way to view this is https://github.com/civicrm/civicrm-packages/pull/212/files?w=1


seamuslee001:
Reviewed this and its all whitespace only changes giving merge on pass

PR #215, opened 2018-06-21
(dev/core#174) Import PSR-16 compliance test

totten:
Notes:
* I've included with the file explain what's changed from upstream.
* This is a compliance test for a published standard. It's not used at runtime. I don't expect to need regular upgrading.


eileenmcnaughton:
Seems fine - no 'live' impact

PR #216, opened 2018-07-07
Issue 243: Authenticate with Backdrop in kcfinder

herbdool:
Addresses https://lab.civicrm.org/dev/core/issues/243

### DIFF
--- a/kcfinder/integration/civicrm.php
+++ b/kcfinder/integration/civicrm.php
@@ -57,6 +57,9 @@ function checkAuthentication() {
     case 'Drupal6':
       $auth_function = 'authenticate_drupal';
       break;
+    case 'Backdrop':
+      $auth_function = 'authenticate_backdrop';
+      break;
     case 'Joomla':
       $auth_function = 'authenticate_joomla';
       break;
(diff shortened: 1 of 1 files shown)


xurizaemon:
At a glance this looks like the right fix for Backdrop, but we really have to keep on the radar eliminating all entry points that aren't the supported CMS authentication. Duplicating URL, path and CMS/auth resolution code through a handful of other endpoints is not sustainable.

herbdool:
@xurizaemon sounds like a good follow-up issue. Or is there one already?

xurizaemon:
I feel like there was one but I could never find it in JIRA. May suit https://lab.civicrm.org/dev/cloud-native/issues

herbdool:
I think the follow-up is a general problem not just for cloud native. It needs to be standardized on the CMS classes.

xurizaemon:
I agree it's a general problem - but I'm using cloud-native as a best practice wishlist here, rather than dropping it on the general issue list.

Totally happy for the issue to be moved or duplicated from where it is now @ https://lab.civicrm.org/dev/cloud-native/issues/16

herbdool:
Sounds good. Let's leave it there. So I'm not sure if someone else is able to test this bug fix or not. It's pretty much identical to the Drupal function but it also checks for the regular location for modules in /modules/ and not just /sites/*/.

kcfinder/integration/civicrm.php
```diff
@@ -108,6 +111,39 @@ function authenticate_drupal($config) {
...
+  $pos = strpos($base_url, '/sites/');
+  if ($pos === FALSE) {
+    $pos = strpos($base_url, '/profiles/');
+  }
+  if ($pos === FALSE) {
+    $pos = strpos($base_url, '/modules/');
+  }
+  $base_url = substr($base_url, 0, $pos); // Backdrop root absolute url
+
+  CRM_Utils_System::loadBootStrap(CRM_Core_DAO::$_nullArray, true, false);
```
eileenmcnaughton: looks like we can just pass an empty array without having to use the static. Note that in rare cases CRM_Core_DAO::$_nullArray can become contaminated.

eileenmcnaughton:
So my feeling is we might struggle to get reviewers on this but 
a) it's fixing a hard error so
b) it would only be worse than existing if it introduced a security flaw - from what I can tell it is equivalent to the drupal function so would not be 'worse' and there is a permission check. I'm going to merge this 

Obviously I agree with @xurizaemon on really wanting to get rid of it - although I tried & failed not that long ago https://github.com/civicrm/civicrm-core/compare/master...eileenmcnaughton:ck?expand=1

laryn:
I just tested with Backdrop 1.10.1 and CiviCRM 5.4.0 and it works.

eileenmcnaughton:
Thanks @laryn

laryn:
@eileenmcnaughton Quick followup to make sure I was clear, as I notice this was merged some time back. The patch is not included in the latest Backdrop download and I had to apply the patch manually. Is that expected?

eileenmcnaughton:
@laryn looks like I merged on Jul 8 - that is after 1st Wed of July so it would have gone into the rc cut on first Wed of Aug - so try the 5.5 rc from https://download.civicrm.org/latest

laryn:
Ah, yes. Thanks!!

PR #220, opened 2018-07-19
"Only variable references should be returned by reference" notice in Mail_smtp

mfb:
Our CI system was misconfigured to try to send mail via SMTP, which helped me uncover a bug in `Mail_smtp->getSMTPObject()`

This should resolve an "Only variable references should be returned by reference" PHP notice in various SMTP error conditions.

Note: Possibly this entire library should be upgraded to current version.


mfb:
jenkins test this please

MikeyMJCO:
I'd be more inclined to test updating the entire library rather than patching it. Have you tested upgrading the library?

mfb:
Pear Mail 1.2.0 was for "PHP 4.4.9 or newer" so that might explain this unnecessary code?

mfb:
Looks like this library has been hacked with CiviCRM-specific code, so there would be some technical debt to deal with while upgrading it :)

mfb:
Is the proper way to upgrade packages to move them into the civicrm-core composer.json file?  Or upgrade in place here?

I will probably open separate PR(s) for that in case someone wants to merge this minor bugfix before that's ready.

totten:
> Is the proper way to upgrade packages to move them into the civicrm-core composer.json file? Or upgrade in place here?

Ideally, yes.

For background, the original managers of `packages` followed this protocol:

1. Copy in the code into `packages/Foo`.
2. If they need to patch the code further, then:
    1. Copy the original code to `packages/_ORIGINAL_/Foo`
    2. Edit the code in `packages/Foo`

So the protocol I've used for upgrading a PHP package like this is:

* Figure out what version and which (if any) patches are currently in `packages`.
    * Hopefully, that's just diff'ing the `Foo` and `_ORIGINAL_/Foo`.
    * Perhaps do `git log` on `Foo` to verify that there weren't any other changes (where someone forgot to copy to `_ORIGINAL`).
    * If you want to be really paranoid, figure the version of `Foo` and compare the canonical release with the code in `packages`.
* Assess the target (upgrade) version
    * Figure out if the local patches should be discarded or brought forward. (The patch might deal with something that's become irrelevant or that's resolved by upstream.)
    * Check upstream release notes for compatibility breaks.
* Prepare patch/PR to remove from `civicrm-packages`
* Prepare patch/PR to add to `civicrm-core/composer.json`
    * If everything's clean, then you can just add the dependency and be done.
    * But sometimes you also need to update `composer.json` with a  `post-install-cmd`/`post-update-cmd`. This can be used to:
        * Apply patches
        * Delete example files (which are potential security vulnerabilities).
* Verify that we use the package via PHP autoloader. (Ex: `new Mail_smtp()` is OK, but `require_once 'packages/Mail/smtp.php'` can be problematic.)

(Note: The protocol for JS/CSS library tends to be different. PHP is a bit better off because we mostly rely on autoloaders.)

mfb:
Looks like some of the patches are no longer necessary. And the rest should be revisited and refactored; e.g. logging should not be added by hacking the library.

mfb:
Oops this one fell off my radar :/  hope to dive back into it soon

mfb:
Ok, here I try to remove pear/mail altogether, in favor of installing it via composer.json in civicrm/civicrm-core#13289 

I'm not sure this will work, because now PEAR.php is available in two different places - vendor and packages.  We might have to move all the other Pear stuff into vendor too?

mfb:
ok I removed the redundant PEAR.php, Console/Getopt.php, etc. from packages, and things seem to still be working fine in my quick testing.

seamuslee001:
Jenkins re test this please

seamuslee001:
Merging as per the tag

PR #221, opened 2018-07-21
(NFC) Remove deprecated function use

colemanw:
Overview
----

Swaps around some functions introduced by @eileenmcnaughton and I for the non-deprecated equivalents. As @totten noted there was some redundancy, so as soon as this is merged I'll delete the other functions from core to keep things tidy.


eileenmcnaughton:
@colemanw I agree these changes make sense and swap out equivalent functions.

I kind of feel like the new functions names are better than the old so I'm a soft +1 here - gave it merge ready rather than merged

seamuslee001:
Merging as per the tag

colemanw:
Thanks @seamuslee001 and @eileenmcnaughton - I've followed up with https://github.com/civicrm/civicrm-core/pull/12539

PR #222, opened 2018-07-22
Update to Quickform Latest version

seamuslee001:
ping @colemanw @eileenmcnaughton @totten this updates to reference the latest quickform version. The diff between it and what we have (including this PR) is seen here https://gist.github.com/seamuslee001/0674a976c4526d2e8abf8a066546a489. There are some changes there that we may want to have but i'm not sure what the scope of them / implications are

### DIFF
--- a/HTML/QuickForm.php
+++ b/HTML/QuickForm.php
@@ -17,9 +17,9 @@
  * @author      Adam Daniel <adaniel1@eesus.jnj.com>
  * @author      Bertrand Mansion <bmansion@mamasam.com>
  * @author      Alexey Borzov <avb@php.net>
- * @copyright   2001-2009 The PHP Group
+ * @copyright   2001-2011 The PHP Group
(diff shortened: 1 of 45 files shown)


colemanw:
FYI @seamuslee001 there are a *lot* of undocumented in-house patches we've made to our copy of quickform over the years. I think if we're serious about this, we'll need to start by doing a diff between our copy and the original of the same vesion, generate a patch, and then upgrade quickform and try to apply the patch.

seamuslee001:
@civicrm-builder re test this please

seamuslee001:
@colemanw that is what i see the gist being part of, its showing the differences between the base and our patched version. Most of the changes are obvious (e.g. us switching to main __constructor style calls) the heiraselect ones i'm not sure there is clearly some hacks about ids etc. Also it shows that core didn't go as far as we did getting rid of the evals

seamuslee001:
closing as replaced with https://github.com/civicrm/civicrm-packages/pull/224

PR #224, opened 2018-07-27
Upgrade to Quickform 3.2.16 

seamuslee001:
This is a revised PR following discussion between myself and @colemanw. Following Coleman's idea i did a diff between the original version 3.2.11 and 3.2.16 and created this gist https://gist.github.com/seamuslee001/0a67ad08359f46b183a7e26f05c99658 of the difference. I left out the changes for security. I should also note that our version of hierselect.php is really really really old. I have therefor left it out of this PR as we need to consider it separately

### DIFF
--- a/HTML/QuickForm.php
+++ b/HTML/QuickForm.php
@@ -17,9 +17,9 @@
  * @author      Adam Daniel <adaniel1@eesus.jnj.com>
  * @author      Bertrand Mansion <bmansion@mamasam.com>
  * @author      Alexey Borzov <avb@php.net>
- * @copyright   2001-2009 The PHP Group
+ * @copyright   2001-2011 The PHP Group
  * @license     http://www.php.net/license/3_01.txt PHP License 3.01
- * @version     CVS: $Id: QuickForm.php,v 1.166 2009/04/04 21:34:02 avb Exp $
+ * @version     CVS: $Id$
  * @link        http://pear.php.net/package/HTML_QuickForm
  */
 
@@ -31,6 +31,10 @@ require_once 'PEAR.php';
  * Base class for all HTML classes
  */
 require_once 'HTML/Common.php';
+/**
+ * Static utility methods
+ */
+require_once 'HTML/QuickForm/utils.php';
 
 /**
  * Element types known to HTML_QuickForm
@@ -120,7 +124,7 @@ define('QUICKFORM_INVALID_DATASOURCE',     -9);
  * @author      Adam Daniel <adaniel1@eesus.jnj.com>
  * @author      Bertrand Mansion <bmansion@mamasam.com>
  * @author      Alexey Borzov <avb@php.net>
- * @version     Release: 3.2.11
+ * @version     Release: 3.2.16
  */
 class HTML_QuickForm extends HTML_Common
 {
@@ -2054,7 +2058,7 @@ class HTML_QuickForm extends HTML_Common
  * @package     HTML_QuickForm
  * @author      Adam Daniel <adaniel1@eesus.jnj.com>
  * @author      Bertrand Mansion <bmansion@mamasam.com>
- * @version     Release: 3.2.11
+ * @version     Release: 3.2.16
  */
 class HTML_QuickForm_Error extends PEAR_Error {
 
--- a/HTML/QuickForm/Rule/Range.php
+++ b/HTML/QuickForm/Rule/Range.php
@@ -15,9 +15,9 @@
  * @category    HTML
  * @package     HTML_QuickForm
  * @author      Bertrand Mansion <bmansion@mamasam.com>
- * @copyright   2001-2009 The PHP Group
+ * @copyright   2001-2011 The PHP Group
  * @license     http://www.php.net/license/3_01.txt PHP License 3.01
- * @version     CVS: $Id: Range.php,v 1.8 2009/04/04 21:34:04 avb Exp $
+ * @version     CVS: $Id$
  * @link        http://pear.php.net/package/HTML_QuickForm
  */
 
@@ -32,7 +32,7 @@ require_once 'HTML/QuickForm/Rule.php';
  * @category    HTML
  * @package     HTML_QuickForm
  * @author      Bertrand Mansion <bmansion@mamasam.com>
- * @version     Release: 3.2.11
+ * @version     Release: 3.2.16
  * @since       3.2
  */
 class HTML_QuickForm_Rule_Range extends HTML_QuickForm_Rule
@@ -45,7 +45,7 @@ class HTML_QuickForm_Rule_Range extends HTML_QuickForm_Rule
      * @access    public
      * @return    boolean   true if value is valid
      */
-    function validate($value, $options)
+    function validate($value, $options = null)
     {
         $length = strlen($value);
         switch ($this->name) {
--- a/HTML/QuickForm/autocomplete.php
+++ b/HTML/QuickForm/autocomplete.php
@@ -15,9 +15,9 @@
  * @category    HTML
  * @package     HTML_QuickForm
  * @author      Matteo Di Giovinazzo <matteodg@infinito.it>
- * @copyright   2001-2009 The PHP Group
+ * @copyright   2001-2011 The PHP Group
  * @license     http://www.php.net/license/3_01.txt PHP License 3.01
- * @version     CVS: $Id: autocomplete.php,v 1.8 2009/04/04 21:34:02 avb Exp $
+ * @version     CVS: $Id$
  * @link        http://pear.php.net/package/HTML_QuickForm
  */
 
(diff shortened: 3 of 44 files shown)


seamuslee001:
Jenkins re-test this please

HTML/QuickForm.php
```diff
@@ -31,6 +31,10 @@
  * Base class for all HTML classes
  */
 require_once 'HTML/Common.php';
+/**
+ * Static utility methods
+ */
+require_once 'HTML/QuickForm/utils.php';
```
eileenmcnaughton: Since most of this is non-change I'm just gonna add comments on the lines that are code changes in the first pass & analyse on the second 

Change 1

HTML/QuickForm/Rule/Range.php
```diff
@@ -45,7 +45,7 @@ class HTML_QuickForm_Rule_Range extends HTML_QuickForm_Rule
      * @access    public
      * @return    boolean   true if value is valid
      */
-    function validate($value, $options)
+    function validate($value, $options = null)
```
eileenmcnaughton: change 2

HTML/QuickForm/autocomplete.php
```diff
@@ -124,7 +124,7 @@ function toHtml()
         $arrayName = str_replace(array('[', ']'), array('__', ''), $this->getName()) . '_values';
 
         $this->updateAttributes(array(
-            'onkeypress' => 'return autocomplete(this, event, ' . $arrayName . ');'
+            'onkeypress' => 'return window.autocomplete(this, event, ' . $arrayName . ');'
```
eileenmcnaughton: change 3

HTML/QuickForm/date.php
```diff
@@ -55,7 +55,7 @@ class HTML_QuickForm_date extends HTML_QuickForm_group
     var $_options = array(
         'format'           => 'dMY',
         'minYear'          => 2001,
-        'maxYear'          => 2012,
+        'maxYear'          => null, // set in the constructor
```
eileenmcnaughton: change 4

HTML/QuickForm/date.php
```diff
@@ -129,6 +129,10 @@ function __construct($elementName = null, $elementLabel = null, $options = array
         $this->_persistantFreeze = true;
         $this->_appendName = true;
         $this->_type = 'date';
+
```
eileenmcnaughton: change 5

HTML/QuickForm/element.php
```diff
@@ -17,16 +17,20 @@
...
  */
 
 /**
  * Base class for all HTML classes
  */
 require_once 'HTML/Common.php';
+/**
+ * Static utility methods
+ */
+require_once 'HTML/QuickForm/utils.php';
```
eileenmcnaughton: change 6

HTML/QuickForm/file.php
```diff
@@ -320,10 +320,12 @@ static function _ruleCheckFileName($elementValue, $regex)
     * Needs to be redefined here as $_FILES is populated differently from
     * other arrays when element name is of the form foo[bar]
     *
+    * @param bool $sc1   unused, for signature compatibility
+    *
     * @access    private
     * @return    mixed
     */
-    function _findValue(&$values)
+    function _findValue(&$sc1 = null)
```
eileenmcnaughton: change 7

HTML/QuickForm/header.php
```diff
@@ -60,10 +60,12 @@ function __construct($elementName = null, $text = null)
     * Accepts a renderer
     *
     * @param HTML_QuickForm_Renderer    renderer object
+    * @param bool $sc1                  unused, for signature compatibility
+    * @param bool $sc2                  unused, for signature compatibility
     * @access public
     * @return void
     */
-    function accept(&$renderer)
+    function accept(&$renderer, $sc1 = false, $sc2 = null)
```
eileenmcnaughton: change 8

HTML/QuickForm/hiddenselect.php
```diff
@@ -105,10 +105,14 @@ function toHtml()
     // }}}
     // {{{ accept()
 
-   /**
-    * This is essentially a hidden element and should be rendered as one
-    */
-    function accept(&$renderer)
+    /**
```
eileenmcnaughton: change 9

HTML/QuickForm/html.php
```diff
@@ -63,10 +63,12 @@ function __construct($text = null)
     * Accepts a renderer
     *
     * @param HTML_QuickForm_Renderer    renderer object (only works with Default renderer!)
+    * @param bool $sc1                  unused, for signature compatibility
+    * @param bool $sc2                  unused, for signature compatibility
     * @access public
     * @return void
     */
-    function accept(&$renderer)
+    function accept(&$renderer, $sc1 = false, $sc2 = null)
```
eileenmcnaughton: change 10

HTML/QuickForm/utils.php
```diff
@@ -0,0 +1,159 @@
+<?php
```
eileenmcnaughton: change 13....

HTML/QuickForm/utils.php
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * LICENSE: This source file is subject to version 3.01 of the PHP license
+ * that is available through the world-wide-web at the following URI:
+ * http://www.php.net/license/3_01.txt If you did not receive a copy of
+ * the PHP License and are unable to obtain it through the web, please
+ * send a note to license@php.net so we can mail you a copy immediately.
+ *
+ * @category    HTML
+ * @package     HTML_QuickForm
+ * @author      Chuck Burgess <ashnazg@php.net>
```
eileenmcnaughton: I read this as Chris Burgess first pass :-)

HTML/QuickForm/Rule/Range.php
eileenmcnaughton: So this is weird - it doesn't match the parent signature - but it didn't before either - feels smelly but I guess....


```
   /**
    * Validates a value
    *
    * @access public
    * @abstract
    */
    function validate($value)
    {
        return true;
    }

```

HTML/QuickForm/file.php
eileenmcnaughton: Did we miss a trick here - it seems less compatible with the parent

```
    /**
     * Tries to find the element value from the values array
     *
     * @since     2.7
     * @access    private
     * @return    mixed
     */
    function _findValue(&$values)
    {
```

HTML/QuickForm/header.php
eileenmcnaughton: Also not matching the parent
```
    * @param HTML_QuickForm_Renderer    renderer object
    * @param bool                       Whether an element is required
    * @param string                     An error message associated with an element
    * @access public
    * @return void
    */
    function accept(&$renderer, $required=false, $error=null)
    {
        $renderer->renderElement($this, $required, $error);
    } // end func accept

    // }}}
```

HTML/QuickForm/file.php
seamuslee001: Not sure that is the change but this is the direct comparison from 3.2.11 to 3.2.16 of the upstream https://gist.github.com/seamuslee001/0a67ad08359f46b183a7e26f05c99658#file-quickform_diff-diff-L748

HTML/QuickForm/hiddenselect.php
eileenmcnaughton: parent is ```
    function accept(&$renderer, $required=false, $error=null)
    {
        $renderer->renderElement($this, $required, $error);
    } // end func accept

    // }}}
```

HTML/QuickForm/html.php
eileenmcnaughton: ditto sig

eileenmcnaughton: ah this might BE the parent!

HTML/QuickForm/header.php
seamuslee001: Again change comes from straight upstream https://gist.github.com/seamuslee001/0a67ad08359f46b183a7e26f05c99658#file-quickform_diff-diff-L812

HTML/QuickForm/html.php
eileenmcnaughton: no it's HTML_QuickForm_element

HTML/QuickForm/hiddenselect.php
seamuslee001: https://gist.github.com/seamuslee001/0a67ad08359f46b183a7e26f05c99658#file-quickform_diff-diff-L851 change comes directly from upstream

eileenmcnaughton:
@seamuslee001 so there is  no chance we've missed our parent signature changes somehow.

The only thing I see is some possible function signature discrepancies - but they won't be quiet failures so I'm ok with merging as we can fix any resulting notices if we need to

seamuslee001:
Thanks @eileenmcnaughton

PR #225, opened 2018-08-03
dev/core#292 Fix issue where accented characters were not properly ut…

seamuslee001:
…f8 encoded which then broke the json_encoding

ping @eileenmcnaughton @francescbassas 

Local Testing suggests this will work, i'm not sure why its not already properly utf8.

### DIFF
--- a/HTML/QuickForm/hierselect.php
+++ b/HTML/QuickForm/hierselect.php
@@ -364,6 +364,7 @@ class HTML_QuickForm_hierselect extends HTML_QuickForm_group
             if ($js != '') {
                 $js .= ",\n";
             }
+            $options = utf8_encode($options);
             $js .= '"'.$optValue.'":'.json_encode($options);
         }
     }


eileenmcnaughton:
Thanks @seamuslee001 great work. @francescbassas does this work for you?

@seamuslee001 I think we should merge to 5.5 & possibly backport to 5.4

seamuslee001:
@eileenmcnaughton did you see my comments on the backstory to this one. Also ping @agileware-fj Francis, this relates to a fix you did for AUG i think where by `\`s were causing issues in the javascript in one part.

francescbassas:
I can confirm that this works for me! Thanks @seamuslee001 and @eileenmcnaughton.

eileenmcnaughton:
@seamuslee001 if you do against the rc I'll merge. That way we can put a patch version out if we do decide it was a 5.3.2 regression & there is not really a downside

seamuslee001:
@eileenmcnaughton i have put this now against the 5.5 branch which is the RC now. I guess question is do we want to do a 5.4.1 release

eileenmcnaughton:
Thanks @seamuslee001 - merged - I think we likely should - if we agree it was a 5.3.2 regression - but perhaps not rush it out in case something else emerges in the next few days? Should we merge a PR to 5.4 so it's there is we do cut one?

(will up-merge now)

eileenmcnaughton:
ok - we are still playing whackamole here - new breakage https://lab.civicrm.org/dev/core/issues/508

PR #227, opened 2018-08-31
Update gettext to latest tagged release

borisson:
screen shot 2018-08-31 at 15 12 42

This improves performance _for free_. By quite a lot if I understand blackfire's results correctly.

This should be tested very well before merging.


mlutfy:
jenkins, test this please

mlutfy:
add to whitelist

borisson:
Yeah, I also found that page earlier, I think that this also provides value though. 

I'm not sure if any of the other libraries need to be updated as well, for better performance/features. But this one stuck out in the blackfire profile.

PHPgettext/gettext.php
```diff
@@ -98,7 +98,7 @@ function readintarray($count) {
    * @param object Reader the StreamReader object
    * @param boolean enable_cache Enable or disable caching of strings (default on)
    */
-  function __construct($Reader, $enable_cache = true) {
+  function gettext_reader($Reader, $enable_cache = true) {
```
seamuslee001: @borisson This should be reversed and can you change file permissions back to 664 pls

PHPgettext/streams.php
```diff
@@ -49,7 +49,7 @@ class StringReader {
   var $_pos;
   var $_str;
 
-  function __construct($str='') {
+  function StringReader($str='') {
```
seamuslee001: @borisson @mlutfy the changes in this file should be reverted (php7 and more so later versions of php5 prefer the __construct style for class constructors

seamuslee001:
@borisson @mlutfy PR looks ok to me, added some comments around things that should be changed. I don't have a multilingual environment to test this on tho

mlutfy:
@borisson Can you remove the `examples` folder? They include some binary `mo` files (small, but still, may as well avoid it).

borisson:
All requested changes have been done.

totten:
jenkins, test this please

borisson:
Looks like this is failing tests, but I don't understand why. They don't seem to be related.

borisson:
As another option, we could remove the library from this package, and require the https://github.com/phpmyadmin/motranslator/ library trough composer instead. That removes maintenance burden from the CiviCRM maintainers.

It's also a better-tested and more modern version of this package.

I'm not sure what the BC policy about that would be, but I think that this is probably the best solution in the future.

seamuslee001:
Jenkins re test this please

seamuslee001:
This looks fine to me now

PR #228, opened 2018-11-06
Fix continue / break warning under PHP 7.3

ejegg:
PHP 7.3 has started to issue warnings when a continue statement is
inside a switch and interpreted as a break
  "continue" targeting switch is equivalent to "break".
  Did you mean to use "continue 2"?
In this case, it looks like 'continue 2' is the right way to go.

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -3767,15 +3767,13 @@ class DB_DataObject extends DB_DataObject_Overload
             // at this point if you have set something to an object, and it's not expected
             // the Validate will probably break!!... - rightly so! (your design is broken,
             // so issuing a runtime error like PEAR_Error is probably not appropriate..
-
-            switch (true) {
-                // todo: date time.....
-                case  ($val & DB_DATAOBJECT_STR):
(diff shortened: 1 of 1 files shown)


ejegg:
strange, api_v3_JobProcessMailingTest::testConcurrency.testConcurrency is failing with all its datasets, but it's hard to tell what this commit could have done. The other 1000+ tests are passing, though!

eileenmcnaughton:
@ejegg that test had a known issue after a server rebuild....

I agree with your analysis on continue 2 being correct. It's a dog of a construct though!
```
if ($val & DB_DATAOBJECT_STR) {
    $ret[$key] = Validate::string($this->$key, VALIDATE_PUNCTUATION . VALIDATE_NAME);
    continue;
}

if (($val & DB_DATAOBJECT_INT):
   $ret[$key] = Validate::number($this->$key, array('decimal'=>'.'));
   continue;
}

```

would save some braincells IMHO

eileenmcnaughton:
test this please

seamuslee001:
test this please

eileenmcnaughton:
merging per previous discussion on this code (above)

PR #229, opened 2018-11-14
Php 7.2 fix notice on countable

eileenmcnaughton:
Fixes php 7.2 enotice by checking for NULL before attempting to count 

screenshot 2018-11-15 11 29 06


colemanw:
Test failure unrelated. Looks like a no-brainer- you can't count NULL.

PR #232, opened 2018-11-19
Fix PHP7.2 warning when re-generating schema files

seamuslee001:
This just fixes an issue when running either ./bin/setup.sh -Dg or ./bin/regen.sh where a warning would be displayed in PHP7.2

### DIFF
--- a/PHP/Beautifier.php
+++ b/PHP/Beautifier.php
@@ -1190,7 +1190,7 @@ class PHP_Beautifier implements PHP_Beautifier_Interface
      */
     public function decIndent()
     {
-        if (count($this->aIndentStack) > 1) {
+        if (count($this->aIndentStack) >= 1) {
             $iLastIndent = array_pop($this->aIndentStack);
             $this->iIndent-= $iLastIndent;
         }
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@seamuslee001 this looks more correct with your change.

However, I get a small style regression when running GenCode after applying it

```
diff --git a/CRM/Core/I18n/SchemaStructure.php b/CRM/Core/I18n/SchemaStructure.php
index 2c7f442856..fc9ccafc03 100644
--- a/CRM/Core/I18n/SchemaStructure.php
+++ b/CRM/Core/I18n/SchemaStructure.php
@@ -710,5 +710,5 @@ class CRM_Core_I18n_SchemaStructure {
     }
     return $result;
   }
-
-}
+  }
+  

```

seamuslee001:
@eileenmcnaughton just pushing a fix for that now i think, not sure how this function ever actually worked sensibly

eileenmcnaughton:
@seamuslee001 agree this results in no change. Adding merge on pass but I'm not sure what to make of the test fails

seamuslee001:
Jenkins re test this please

PR #239, opened 2019-02-07
Upgrade Jquery contained within KcFinder

seamuslee001:
This upgrades the jQuery version within KcFiner to a stable version

@colemanw @totten @eileenmcnaughton


eileenmcnaughton:
I had mixed results with & without this patch - I'm gonna merge it so it hits the rc & gets the most possible testing

kcfinder is a 'mare

PR #240, opened 2019-02-13
Remove old menu plugin

colemanw:
Followup to https://github.com/civicrm/civicrm-core/pull/13582
These files can be removed once that is merged.


seamuslee001:
@colemanw should this be merged now?

PR #241, opened 2019-02-18
Change connection and results charsets to utf8mb4 on MySQL 5.5.3+

mfb:
CiviCRM isn't using utf8mb4 yet, but let's start the process of migrating to utf8mb4... We can use utf8mb4 as the connection and results charset on MySQL 5.5.3+


colemanw:
@mlutfy can you review?

mfb:
@totten this PR is essentially the same as https://github.com/civicrm/civicrm-setup/pull/16 so I think it could be a candidate for being merged (the version logic is in the query itself rather than in PHP code).

totten:
@mfb yup, agree that looks sensible.

totten:
jenkins, test this please

totten:
It's also good to get this in at the start of the dev-cycle, which would be now. :)

Adding merge-on-pass. Small proviso: after the test-suite passes, but before merging, just do a quick login on the autobuild test site.

PR #243, opened 2019-03-16
remove amavisd

MegaphoneJon:
I was surprised to see amavisd-new in the packages directory.  I researched and apparently in Civi 1.x and 2.x it was used to send mail.  I've been using Civi for 9 years and never even knew about this configuration; I can't see any reference to it in the codebase (except a minor one in distmaker) and can't see how it might still be used today.


seamuslee001:
I'm going to add merge on pass on this on the basis that this directory is actually excluded when building the distributions so only really need to see if unit tests still pass with it

PR #245, opened 2019-03-16
Remove Log.php from the Original folder as well

seamuslee001:
This removes the Log.php folder from original folder as well which is just there as a backup copy no longer needed


eileenmcnaughton:
unrelate fails IMHO -  merging

PR #246, opened 2019-03-17
Remove htmlpurifier from within the IDS to facilitate it being provid…

seamuslee001:
…ed by composer

This removes the html purifier from within the IDS as it will be provided by composer as per https://github.com/civicrm/civicrm-core/pull/13840 ping @totten @mfb


IDS/Config/Config.ini.php
```diff
@@ -19,8 +19,8 @@
     
     ; in case you want to use a different HTMLPurifier source, specify it here
     ; By default, those files are used that are being shipped with PHPIDS
-    HTML_Purifier_Path	= vendors/htmlpurifier/HTMLPurifier.auto.php
-    HTML_Purifier_Cache = vendors/htmlpurifier/HTMLPurifier/DefinitionCache/Serializer
+    HTML_Purifier_Path	= ../../../vendor/ezyang/htmlpurifier/library/HTMLPurifier.auto.php
```
seamuslee001: This one maybe problematic especially on a d8 build but not 100% sure where packages go in d8 builds

IDS/Config/Config.ini.php
```diff
@@ -19,8 +19,8 @@
     
     ; in case you want to use a different HTMLPurifier source, specify it here
     ; By default, those files are used that are being shipped with PHPIDS
-    HTML_Purifier_Path	= vendors/htmlpurifier/HTMLPurifier.auto.php
-    HTML_Purifier_Cache = vendors/htmlpurifier/HTMLPurifier/DefinitionCache/Serializer
+    HTML_Purifier_Path	= [civicrm.root]/vendor/ezyang/htmlpurifier/library/HTMLPurifier.auto.php
+    HTML_Purifier_Cache = [civicrm.root]/vendor/ezyang/htmlpurifier/library/HTMLPurifier/DefinitionCache/Serializer
```
seamuslee001: @totten @mfb note that because i am unsure of the directory structure for within D8 i have aimed to go with a relative path here and patched up the IDS/Monitor.php to support the relative path

mfb:
by the way, it looks like at some point the HTMLPurifier cache ended up at packages/IDS/tmp - these files don't need to be here.

seamuslee001:
looks like its meant to be in the searlizer folder https://github.com/PHPIDS/PHPIDS/blob/master/lib/IDS/Config/Config.ini.php#L21 likewise to what i am changing above

mfb:
Ah well those should go too - we need to cache to a writable directory.

I think hypothetically you can distribute the cache files - if there is no configurability - and this may have been the initial idea.  But those cache files would need to be regenerated for each version of HTMLPurifier, as they are version-specific.

seamuslee001:
Jenkins re test this please

PR #248, opened 2019-04-19
Extend CRM-21489 handling to ensure that all deadlock errors are trea…

seamuslee001:
…ted as Exceptions rather than maybe just sometimes Pear Error objects and sometimes Exceptions

### DIFF
--- a/DB/mysqli.php
+++ b/DB/mysqli.php
@@ -928,9 +928,16 @@ class DB_mysqli extends DB_common
             }
             $errno = $this->errorCode(mysqli_errno($this->connection));
         }
-        return $this->raiseError($errno, null, null, null,
+        $errorObject = $this->raiseError($errno, null, null, null,
                                  @mysqli_errno($this->connection) . ' ** ' .
                                  @mysqli_error($this->connection));
+        // If we are dealing in a Deadlock then pass this to the Exception Handler so that
+        // It can be caught by the Exception handling in CRM-21489
+        $deadlockIds = [1213, 1205];
+        if (in_array(@mysqli_errno($this->connection), $deadlockIds)) {
+            CRM_Core_Error::exceptionHandler($errorObject);
+        }
+        ;return $errorObject;
     }
 
     // }}}
--- a/DB/mysql.php
+++ b/DB/mysql.php
@@ -890,9 +890,16 @@ class DB_mysql extends DB_common
             }
             $errno = $this->errorCode(mysql_errno($this->connection));
         }
-        return $this->raiseError($errno, null, null, null,
+        $errorObject = $this->raiseError($errno, null, null, null,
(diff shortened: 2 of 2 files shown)


eileenmcnaughton:
@seamuslee001 I had a go at replicating this since I got around to doing a PR for https://github.com/civicrm/civicrm-packages/pull/258 & updating https://github.com/civicrm/civicrm-packages/pull/180 with a test & I had partial success but I concluded this was not correct.

In my testing I created a table

```
CREATE TABLE `innodb_deadlock_maker` (
  `a` int(11) NOT NULL,
  PRIMARY KEY (`a`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8
```

In a mysql console I typed ``` BEGIN```

I then put these 2 rows into a unit test (a random one) 

```
CRM_Core_DAO::executeQuery('BEGIN');
CRM_Core_DAO::executeQuery('select * from test.innodb_deadlock_maker for update');
```

I paused between them so that before running the second line I went back to my mysql window & entered

```
insert into innodb_deadlock_maker values(8);
```

I then let the second row run - resulting in a deadlock. In fact locally I couldn't capture the errno - (which is probably a local issue) but I also found that despite that the Exception was always thrown. I concluded that it was because earlier on I had passed through the line 

```
    $errorScope = \CRM_Core_TemporaryErrorScope::useException();
```

in setup.

As a result I concluded that the reason these deadlocks are inconsistent is probably because we are not setting our error scope to use Exceptions consistently. The comment block on the class has the wise advice 

"This class and all references to it should be removed in 4.3/4.4 (when we adopt exception-based error handling)."

Generally we seem to have exception based error handling in 'most' flows now I think - but drush feels like a possible exception. I think the correct fix for this issue is in fact to determine how exception handling is being bypassed & fix that. As a bonus we'll get cleaner code, more consistent errors and better output (drush still outputs the whole fatal html sometimes :-()

DB/mysqli.php
```diff
@@ -928,9 +928,16 @@ function mysqliRaiseError($errno = null)
...
+        $errorObject = $this->raiseError($errno, null, null, null,
                                  @mysqli_errno($this->connection) . ' ** ' .
                                  @mysqli_error($this->connection));
+        // If we are dealing in a Deadlock then pass this to the Exception Handler so that
+        // It can be caught by the Exception handling in CRM-21489
+        $deadlockIds = [1213, 1205];
+        if (in_array(@mysqli_errno($this->connection), $deadlockIds)) {
+            CRM_Core_Error::exceptionHandler($errorObject);
+        }
+        ;return $errorObject;
```
ejegg: stray semicolon

eileenmcnaughton:
I think this got otherwise resolved (probably why it is stale)

PR #251, opened 2019-05-21
(dev/joomla#9) htmlspecialchars() expects parameter 1 to be string

spalmstr:
This error can occur: htmlspecialchars() expects parameter 1 to be string, array given in .../administrator/components/com_civicrm/civicrm/packages/HTML/Common.php on line 144

The patch test to see if the value is an array.  If it is, it calls htmlspecialchars for each member.  If the array is empty it calls the same function with a null string.

See: [https://lab.civicrm.org/dev/joomla/issues/9](https://lab.civicrm.org/dev/joomla/issues/9)

### DIFF
--- a/HTML/Common.php
+++ b/HTML/Common.php
@@ -134,17 +134,33 @@ class HTML_Common
      * @return   string
      * @access   private
      */
-    function _getAttrString($attributes)
-    {
-        $strAttr = '';
+    function _getAttrString($attributes)    {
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
I'm going to close this because we've found that we have had this notice before & it's an indicator of a problem somewhere else - we've fixed a few places since this was reported & we should get steps to replicate for any current occurrences

PR #257, opened 2019-07-01
expose smarty's compile_check to be overridden in civicrm.settings.php

sluc23:
Related issue [#1073](https://lab.civicrm.org/dev/core/issues/1073).

Smarty variable `compile_check` is hardcoded to **true** by default. This PR exposes it so can be set up to **false** in `civicrm.settings.php` for prod environments, which leads to a performance improvement in page rendering

### DIFF
--- a/Smarty/Smarty.class.php
+++ b/Smarty/Smarty.class.php
@@ -53,6 +53,10 @@ if (!defined('SMARTY_CORE_DIR')) {
     define('SMARTY_CORE_DIR', SMARTY_DIR . 'internals' . DIRECTORY_SEPARATOR);
 }
 
+if (!defined('SMARTY_COMPILE_CHECK')) {
+    define('SMARTY_COMPILE_CHECK', true);
+}
+
 define('SMARTY_PHP_PASSTHRU',   0);
 define('SMARTY_PHP_QUOTE',      1);
 define('SMARTY_PHP_REMOVE',     2);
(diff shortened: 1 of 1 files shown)


sluc23:
please @eileenmcnaughton @seamuslee001 take a look a this. If it's ok this approach, should I create a PR on **civicrm-core**  to add this variable in `civicrm.settings.php` template ?

demeritcowboy:
Rather than a patch against the smarty package itself, could this be done in CRM_Core_Smarty::initialize()? Would that be more maintainable long-term?

eileenmcnaughton:
So in general I agree with the idea that we should be able to alter this setting via a define in civicrm.settings.php on an experimental basis. Later on, when we have adequate feedback, we might choose to make this have a default for production servers that differs from dev servers. To that end I think

1) @demeritcowboy is right - we can do this in CRM_Core_Smarty - that class extends Smarty and we should be able to override in construct or initialize
2) I'd rather see it call a function - ie $this->compile_check = $this->isSmartyCompiled() - allowing us to swap out a setting or prod vs dev assumption later
3) can the text in civicrm.settings.php describe it as experimental & link back to the gitlab - with a specific request that people comment on it with their experiences

sluc23:
moving PR to **civicrm-core** --> https://github.com/civicrm/civicrm-core/pull/14706

sluc23:
closed!

PR #258, opened 2019-07-03
Improve deadlock error handling

eileenmcnaughton:
This simply captures the error when the deadlock is first hit & then if the deadlock threshold it hit, causing an abort, it re-throws the original error - so the backtrace is retained. 

We have been running this in production for about 6 months and the errors have been more meaningful

(@seamuslee001 you might have an interest)


seamuslee001:
Change looks fine to me and unit test have passed so merging

PR #259, opened 2019-07-25
CRM-20455 - DB_common::modifyQuery() - Optimize

totten:
Overview
--------

The recently merged PR #180 added `modifyQuery()` to allow dynamic inspection/modification of most SQL queries (*post-boot*).

This patch modifies and micro-optimizes the new (unreleased) interface.

Ordinarily, one wouldn't care about this level of micro-optimization; however, it's entirely foreseeable that this code will run hundreds or even thousands of times while a user works with a single page.

__Depends__: https://github.com/civicrm/civicrm-core/pull/14882,  https://github.com/civicrm/civicrm-core/pull/14883

Before
------

For each typical SQL query, the call to `modifyQuery()` requires:

* Constructing an instance of `GenericHookEvent`, which does some   array shuffling and encourages reflective code.
* Performing multiple lookups of the container/dispatcher system

After
-----

For each typical SQL query, the call to `modifyQuery()` requires:

* Constructing an instance of `QueryEvent`, which is a lighter object
* Fewer lookups (`isset(Civi::$static[...])`)

Comments
--------

I did some simple benchmarks -- running 10,000 simple SQL queries with various forms of `modifyQuery()`.  The baseline (no events) had the fastest average (`9.444`s).  Adding GHE bumped runtime up +9% (`10.298`s).  With this patch, it only adds +3.5% (`9.787`s).

The averages are based on 5 samples each.  For full details, see https://gist.github.com/totten/da21a8386a7ddcc62def5d78e912bf8c

### DIFF
--- a/DB/common.php
+++ b/DB/common.php
@@ -1148,14 +1148,20 @@ class DB_common extends PEAR
      */
     function modifyQuery($query)
     {
+        // This section of code may run hundreds or thousands of times in a given request.
+        // Consequently, it is micro-optimized to use single lookup in typical case.
+        if (!isset(Civi::$statics['db_common_dispatcher'])) {
+            if (class_exists('Civi\Core\Container') && \Civi\Core\Container::isContainerBooted()) {
+                Civi::$statics['db_common_dispatcher'] = Civi\Core\Container::singleton()->get('dispatcher');
+            }
+            else {
+                return $query;
+            }
+        }
 
-      if (class_exists('Civi\Core\Container') && \Civi\Core\Container::isContainerBooted()) {
-        Civi\Core\Container::singleton()->get('dispatcher')->dispatch('civi.db.query',
-          \Civi\Core\Event\GenericHookEvent::create(array('query' => &$query))
-        );
-      }
-      return $query;
-
+        $e = new \Civi\Core\Event\QueryEvent($query);
+        Civi::$statics['db_common_dispatcher']->dispatch('civi.db.query', $e);
+        return $e->query;
     }
 
     // }}}


eileenmcnaughton:
thanks @totten 300ms over 10k queries seems OK to me & def better!

seamuslee001:
Error was 

```
/home/jenkins/bknix-dfl/build/pkg-259-592jo/web/sites/all/modules/civicrm/packages/DB/common.php).
Error: Class &#039;Civi\Core\Event\QueryEvent&#039; not found in DB_common->modifyQuery() (line 1162 of /home/jenkins/bknix-dfl/build/pkg-259-592jo/web/sites/all/modules/civicrm/packages/DB/common.php).
Drush command terminated abnormally due to an unrecoverable error.       [error]
```
which makes sense as the core PR hadn't been merged

seamuslee001:
Jenkins re test this please

totten:
Cool, the second test-run showed it 14882 fixed the `QueryEvent` error. It also revealed some test-harness issues, which are addressed in another PR: https://github.com/civicrm/civicrm-core/pull/14883

eileenmcnaughton:
test this please

eileenmcnaughton:
Just a note that I'm doing some performance tests before deploying some updates. 

My first tests on 5.13 were importing (via  script) around 462 contacts with contributions per minute. On 5.17 both with and without using this query modifier I was getting around 484 contributions per minute.

I don't consider the above to be reliable enough to say 5.17 is definitely faster than 5.13 but I think it is reliable enough to say the query modifier is not dragging us down

PR #260, opened 2019-07-29
Update "When" package

agileware-pengyi:
Overview
The old package generates wrong dates in some cases.
For example:
setting
result

Before
Wrong recurring dates in some cases.

After
Correct dates with the above settings.

Comments
This update required some changes in the core. see [this PR](https://github.com/civicrm/civicrm-core/pull/14903).
Don't know how to work with the composer here.

Agileware ref: CIVICRM-1223


mattwire:
add to whitelist

totten:
jenkins, add to whitelist

mlutfy:
I notice that the README says PHP 7.1+ ?

agileware-pengyi:
Using composer instead. see https://github.com/civicrm/civicrm-core/pull/15223
Close

PR #261, opened 2019-07-30
(NFC) VERSIONS.php - Add discussion about how to migrate to composer

totten:


seamuslee001:
looks good to me @totten

PR #262, opened 2019-08-12
dev/core#772 fix php7.2 warnings on import

eileenmcnaughton:
Overview
Fixed a notice when importing with php 7.2

This is replicable if you have an import file with the column 'Contact Source' among others & import setting it to
use 'Column Headers'. It correctly maps contact source & sets it as a default. 

Before
Screen Shot 2019-08-12 at 12 00 15 PM

After
No notice

Technical Details
The code in setValue
is open to either a string or an array but it uses count() in it's initial check - which was OK back when it was written
- nowadays count on a string puts out an error (but still returns 1) as per originally

### DIFF
--- a/HTML/QuickForm/hierselect.php
+++ b/HTML/QuickForm/hierselect.php
@@ -259,7 +259,10 @@ class HTML_QuickForm_hierselect extends HTML_QuickForm_group
      */
     function setValue($value)
     {
-        $this->_nbElements = count($value);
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
test this please

jitendrapurohit:
I was able to replicate this and confirm that the patch fixes the error. The "before" notice screen looks like below on contribution import -

image

After - No error on import screen.

eileenmcnaughton:
thanks @jitendrapurohit

PR #266, opened 2019-10-09
INTL_IDNA_VARIANT_2003 deprecated in PHP_7.2

magnolia61:
INTL_IDNA_VARIANT_2003 is deprecated in php 7.2 but the function idn_to_ascii still uses it as default. 

It seems the only way to avoid the warning is calling the function using explicitly the parameter INTL_IDNA_VARIANT_UTS46.


seamuslee001:
Looks right to me

seamuslee001:
test fails are unrleated, i'm going to merge this as it is consistant with the change that is going to happen anyway via https://wiki.php.net/rfc/deprecate-and-remove-intl_idna_variant_2003

PR #267, opened 2019-10-09
Remove open flash chart library now that that we are using d3 for cha…

seamuslee001:
…rts in core

ping @artfulrobot


seamuslee001:
also ping @eileenmcnaughton

PR #268, opened 2019-10-10
Remove Open Flash charts from Version file

seamuslee001:


colemanw:
Hooray no more open flash charts!

PR #271, opened 2019-12-12
dev/core#1469 Mitigate for users that are running ICU 4.4 as the cons…

seamuslee001:
…tant is only defined in ICU 4.6

This mitigates for the possibility where some users their php-intl version is tied to ICU 4.4 or less and the new constant is only available in 4.6


seamuslee001:
this is basically a direct lift of the style from https://forge.typo3.org/projects/typo3cms-core/repository/revisions/0feb1eec075af6ca897dc0f6343507c853aee5f3/diff

eileenmcnaughton:
Fix seems safe & consistent with gitlab discussion

PR #272, opened 2019-12-14
Remove Net_UserAgent_Detect package

seamuslee001:
This removes the Net_UserAgent_Detect package that is no longer supported by Pear https://pear.php.net/package/Net_UserAgent_Detect and i cannot see anywhere that it is called at all

ping @eileenmcnaughton


seamuslee001:
ping @demeritcowboy @mattwire

demeritcowboy:
Since this seems to be a follow-on about wordfence mistakenly flagging APC.php from [stackexchange](https://civicrm.stackexchange.com/questions/34059/packages-net-useragent-detect-apc-php-flagged-as-malicious-by-wordfence-firewall), I'll add that my own searching there suggested the same. It seems to have been originally added as a dependency for paypal a looooonnnnng time ago.

eileenmcnaughton:
OK - I'm prepared to merge off your joint investigation

PR #275, opened 2019-12-18
dev/core#395 Define error messages for deadlock.

eileenmcnaughton:
Overview 
========================

Defines additional error messages for deadlocks. This is fairly deep in the code & not visible to users

Note there is some precedent for this in our codebase per https://issues.civicrm.org/jira/browse/CRM-14571 and I did an upstream PR
https://github.com/pear/DB/pull/9 although I have no idea if it's likely to be reviewed since you know .... old

Before 
========================
When a deadlock is thrown the message winds up as DBError::Unknown Error

After
========================
The error is identified as DBError::deadlock

Screen Shot 2019-12-19 at 9 50 53 AM


eileenmcnaughton:
@seamuslee001 can you take a look at this one - we rolled it out in prod the other day & have been better handling deadlocks in our own code. I also want to fix a core place where there is a misleading error which I can do with this (although I need to pin @totten down to talk about the best place to catch & retry deadlocks cos I think we got it wrong)

seamuslee001:
This works for me, AUG have deployed this to our production and its working well merging

PR #279, opened 2019-12-28
Remove PEAR package Structures_Graph as there appears to be no use fo…

seamuslee001:
…r it

ping @totten @eileenmcnaughton @colemanw


colemanw:
Can't find any reference to it in civicrm core.

PR #276, opened 2019-12-24
Upgrade DB package to be version 1.9.3

seamuslee001:
This upgrades the version of the DB package to be 1.9.3 with our customisations on top

PR #280, opened 2020-01-03
Update DB_Object

eileenmcnaughton:


DB/common.php
```diff
@@ -1161,6 +1161,16 @@ function modifyQuery($query)
         $e = new \Civi\Core\Event\QueryEvent($query);
         Civi::$statics['db_common_dispatcher']->dispatch('civi.db.query', $e);
         return $e->query;
+
```
seamuslee001: @eileenmcnaughton i believe that this was an earlier version of what is in the lines above

DB/common.php
```diff
@@ -2264,6 +2274,9 @@ function _convertNullArrayValuesToEmpty(&$array)
         }
     }
 
+    function lastInsertId() {
```
seamuslee001: I believe this is a duplicate of L2288 (a modification of the patch included in #276

DB/common.php
seamuslee001: see https://github.com/civicrm/civicrm-packages/commit/4483923c7c066f71beae92bd3e5e7d25877538ec#diff-e87e96fe24f44eda22f348fa7adc7246

seamuslee001:
I'm going to treat this as a reviewers cut so merging

eileenmcnaughton:
@seamuslee001 yes I'm happy with that - it was a bit of a slog to do but we both reached the same conclusion so yay

PR #283, opened 2020-01-04
Remove unmaintained unsupported System_Command package

seamuslee001:
The System_Command package does not appear to be used at all in CiviCRM and is now unmaintained in PEAR so lets remove it

ping @eileenmcnaughton @colemanw @totten https://pear.php.net/package/System_Command


eileenmcnaughton:
Yeah - looks like the sort of thing that has security risks too

PR #284, opened 2020-01-26
Remove what appears to be unused SymfonyComponents/YAML and Pear/Date…

seamuslee001:
… Pacakges

I cannot see any references to these 2 packages in core and Symphony/YAML has changed significantly since and Pear/Date is no longer supported

ping @totten @colemanw @eileenmcnaughton


colemanw:
@totten doesn't one of your extensions use YAML?

eileenmcnaughton:
I thought we used yaml in the unit tests

seamuslee001:
if we were still using yaml then i would recommend we use the current symfony/yaml package this yaml class http://www.symfony-project.org/api/1_0/sfYaml.html relates to very old version it appears.

eileenmcnaughton:
Yeah I'm not 100% sure we need i but I'd lean towards updating YAML (via vendor?) & removing PEAR

totten:
> (@colemanw) @totten doesn't one of your extensions use YAML?

Feels unlikely - my general reaction to using YAML is, "I'd love to, but it's not worth creating dep-hell to load `symfony/yaml:^3` across the different deployment styles." I vaguely recall that we noticed this copy of `sfYaml` while we were sprinting on api4/form-builder and considered using it... but I don't think we did.

Grepping `universe`, I can't find anything that uses `sfYaml` (except for some obscure test of a transitive dependency...). So seems safe. IMHO.

> (@eileenmcnaughton) I thought we used yaml in the unit tests

That sounds vaguely familiar, but I think we got rid of it during a round of simplification, and there's no reference to it now.

> (@seamuslee001) if we were still using yaml then i would recommend we use the current symfony/yaml package this yaml class http://www.symfony-project.org/api/1_0/sfYaml.html relates to very old version it appears.

If we really did need a YAML parser, it'd be a lot safer/easier IMHO to use this ancient, obscure implementation to head-off dependency-graph issues.

But as far as I can tell, nothing in the known/published universe uses it.

totten:
FWIW, `civicrm-packages/Validate.php` references `Date_Calc`.

(I haven't been able to find other references to the PEAR Date class, though TBH this is hard to scan.)

seamuslee001:
@totten my argument would be is that we have a number of unit tests that relate to dates and date handlng and they all pass which should give us confidence in removing this pear library tbh.

@eileenmcnaughton is Tim's comment good enough to warrent a merge here?

eileenmcnaughton:
Lol - I was hoping @totten would actually be bold & click merge :-)

totten:
If it was just the YAML, I probably would've. I spooked on all the `Date/*.php` files.

The only confirmed reference I could find is from `Validate.php` (`Valdiate::date()`). I've tried to see if `Validate::date()` might be used. This is tricky -- e.g. the symbols "validate" and "date" are quite common, and the class has some self-referential dynamism via `call_user_func` (in `Validate::multiple()`) that heightens the ambiguity. But overall, the only references to `Validate.php` which I can confirm are in `DB/DataObject.php` (`Validate::string()`, `Validate::number()`), and neither of those should be able to call `Validate::date()`.

I'd give 3/4 odds that it's safe to remove those "Date" classes.

eileenmcnaughton:
@seamuslee001 https://github.com/civicrm/civicrm-packages/blob/46b5d0a9430e5f193c8f1bd470cabe6ed682a00a/DB/DataObject.php#L4592

eileenmcnaughton:
Screen Shot 2020-01-30 at 7 00 37 PM

seamuslee001:
I'm going to close this one given we need the Date package the others were covered in #285

PR #286, opened 2020-02-09
[REF] Use relative path for finding the advmultiseletct javascript

seamuslee001:
ping @totten @eileenmcnaughton


colemanw:
I want to get rid of that thing so bad.

PR #287, opened 2020-02-29
[REF] Remove FPDI library from packages as it is deployed by composer

seamuslee001:
This PR is dependant on https://github.com/civicrm/civicrm-core/pull/16656 

ping @totten @eileenmcnaughton @demeritcowboy


eileenmcnaughton:
Refer to discussion here https://github.com/civicrm/civicrm-core/pull/16656 - we concluded it is not used in core and CDNTaxReceipts  https://github.com/jake-mw/CDNTaxReceipts/tree/master/FPDI has it's own copy (it was originally copied in from that implementation)

PR #292, opened 2020-05-10
Add trigger for smarty error event

mattwire:
See https://github.com/civicrm/civicrm-core/pull/16918

They need to be merged at the same time.

We can't override this function in `CRM_Core_Smarty` because it actually gets triggered from the `Smarty_Compiler` class which is not inherited.

Previously smarty errors (such as parse errors) were silently ignored in most cases making debugging templates extremely difficult.


seamuslee001:
as per civicrm/civicrm-core#16918 confirmed that this works merging

PR #297, opened 2020-06-13
Upgrade HTML Purifier to v4.12.0 to support for PHP7.4

seamuslee001:
ping @demeritcowboy


demeritcowboy:
I clicked on some things here (which is php 7.1) and looked through the code and even the 4.11 functional changes don't seem too scary. I don't have php 7.4 anywhere but this will give me an excuse to do that.

seamuslee001:
Merging this as per @demeritcowboy 's review

PR #299, opened 2020-07-16
[REF] Remove some deprecated size function calls replaced with length

seamuslee001:
as per https://github.com/civicrm/civicrm-core/pull/17871 size() is deprecated and length should be used

ping @colemanw


seamuslee001:
Test fails unrelated

PR #302, opened 2020-08-01
[REF] Remove DB Package as it is now deployed using composer

seamuslee001:
This depends on https://github.com/civicrm/civicrm-core/pull/18027


seamuslee001:
Jenkins re test this please

seamuslee001:
Jenkins re test this please

seamuslee001:
@demeritcowboy @eileenmcnaughton big green tick now

demeritcowboy:
Cool. One thing to note is that for people who manage git installs, the order you do things will matter here. You need to pull packages, then run composer install on the main tree, otherwise the include_path doesn't include vendor/pear/db.

UI seems ok. I tried some different entry points like cron/cv/cli and looks good. Only tried drupal 7.

eileenmcnaughton:
OK as discussed - we wanted to merge this straight after the rc is cut for max pre-release time - so that happened

PR #306, opened 2020-09-05
[REF] Remove Eway Libraries and XML_Util as they are now shipped as p…

seamuslee001:
…art of the representative eway extensions and XML_Util was only used by eway

@eileenmcnaughton once the eway extension PR is merged I think we can do this, you may want to create a new release eventually of the ewayrecurring so its clear people should update


seamuslee001:
Jenkins re test this please

PR #307, opened 2020-09-11
Add hack to support buttonrama without change to code adding submit

eileenmcnaughton:
This is a bit hacky but I looked at cividiscount & concluded we should support the existing
code if we could - if nothing else then to simplify messaging about buttonrama. With this change no change is required
to cividiscount (although I did put up another change to cividiscount which I found to be broken on a different issue https://github.com/civicrm/org.civicrm.module.cividiscount/pull/242)

Without this

Screen Shot 2020-09-11 at 4 53 23 PM

With this

Screen Shot 2020-09-11 at 4 47 42 PM

Having figured this out I'm not 100% sure it was worth the work I put in :-). However, it does mean that no version specific code is required in extensions. I think only a small handle of extensions are affected (including on by @agh1 who I'm sure knows what to do) - but OTOH the way cividiscount does it is nasty & not having to change that right now feels like a bonus

### DIFF
--- a/HTML/QuickForm.php
+++ b/HTML/QuickForm.php
@@ -636,7 +636,26 @@ class HTML_QuickForm extends HTML_Common
            $elementObject->onQuickFormEvent('updateValue', null, $this);
         } else {
             $args = func_get_args();
-            $elementObject =& $this->_loadElement('addElement', $element, array_slice($args, 1));
+            if ($element === 'submit') {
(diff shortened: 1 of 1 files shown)


seamuslee001:
I tested this and confirmed it works with the CiviDiscount extension and will assist with the transition merging

eileenmcnaughton:
thanks @seamuslee001

PR #308, opened 2020-09-14
dev/translation#50 Make recapture library locale aware

seamuslee001:
This picks up the patch from https://civicrm.stackexchange.com/questions/11533/recaptcha-update-language/37503 by @adixon 

@mattwire @mlutfy @MikeyMJCO

### DIFF
--- a/recaptcha/recaptchalib.php
+++ b/recaptcha/recaptchalib.php
@@ -114,7 +114,9 @@ function recaptcha_get_html ($pubkey, $error = null, $use_ssl = false)
         } else {
                 $server = RECAPTCHA_API_SERVER;
         }
-
+        // dev/translation#50 Make sure recapture is locale aware.
+        $locale = CRM_Core_I18n::getLocale();
+        $server = $server .'?hl='.$locale;
         $errorpart = "";
         if ($error) {
            $errorpart = "&amp;error=" . $error;


adixon:
Thank for this. I wasn't convinced this is the right approach - including civi-specific code within a library seems wrong - but least we've now got a conversation. The right solution would be to update the upstream library upstream with a patch that included language in the library call as a parameter, and then include that getLocale call in the civi code that calls the library function. Or maybe there's already a better upstream version? Yeah, that's a whole new can of worms.

MikeyMJCO:
It's the right can of worms imo. We should avoid patching third party libraries and opt for libraries that meet our needs or upstream patches.

MikeyMJCO:
I'm not sure what library we're using but the official PHP lib is supported and language aware... https://github.com/google/ReCAPTCHA

MikeyMJCO:
Definitely in favour of the effort going towards "modern recaptcha" library vs patch ancient unsupported library.

adixon:
So, that leaves us with the work of trying to update the places in civi code that are using this newer version of the library and update the calls to what is almost certainly a more modern and complicated library ...

MikeyMJCO:
Work we're going to have to do anyway eventually... either when PHP makes something in this 2007 code break or when Google turn off RECAPTHA v2. 

I'm going to look at what it would take to switch - the invocation of both is pretty similar and this would give us scope to introduce v3 support which has been requested.

MikeyMJCO:
So the library I linked is more for server-side verification. There is no client side replacement, seems like we'd be better off dropping the library and moving the functions to generate recaptcha elements directly into core.

All the library seems to be doing is inserting the HTML for the recaptcha elements?

seamuslee001:
@MikeyMJCO the library also does the post to run the validation I think as well but this pr will need to be redone once #317 is merged

seamuslee001:
closing as this is now conflicted

PR #309, opened 2020-10-02
Append CIVICRM_DEBUG_LOG_QUERY to sql log var

eileenmcnaughton:
Same as https://github.com/civicrm/civicrm-core/pull/18471 but for the other affected place

@totten


totten:
Looks sensible.

Re-tested with similar procedure as before, but with tighter monitoring of the outputted files, e.g.

```
rm ~/bknix/build/dmaster/web/sites/default/files/civicrm/ConfigAndLog/*log

env CIVICRM_DEBUG_LOG_QUERY=_test_2 cv flush

ls ~/bknix/build/dmaster/web/sites/default/files/civicrm/ConfigAndLog/
```

Confirmed that (1) the `QUERY DONE` statements now appear the alternate query log (`_test_2`) and (2) no other data is written to the default query log.

PR #311, opened 2020-10-29
dev/core#2150 Update checking of response function to work with v2 api

seamuslee001:
This updates the check answer function to work with the v2 API as per https://developers.google.com/recaptcha/docs/verify

### DIFF
--- a/recaptcha/recaptchalib.php
+++ b/recaptcha/recaptchalib.php
@@ -37,7 +37,9 @@
  */
 define("RECAPTCHA_API_SERVER", "http://www.google.com/recaptcha/api.js");
 define("RECAPTCHA_API_SECURE_SERVER", "https://www.google.com/recaptcha/api.js");
-define("RECAPTCHA_VERIFY_SERVER", "www.google.com");
+define("RECAPTCHA_VERIFY_SERVER", "https://www.google.com");
+
+use GuzzleHttp\Client;
 
 /**
  * Encodes the given data into a query string format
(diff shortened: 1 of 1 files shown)


seamuslee001:
@MikeyMJCO @johntwyman @a-cormick-dockery @eileenmcnaughton

johntwyman:
@seamuslee001 I think there's another task to modify [CRM/Utils/ReCAPTCHA.php](https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/ReCAPTCHA.php) in core to make use of this updated library function, right?

recaptcha/recaptchalib.php
```diff
@@ -62,33 +64,17 @@ function _recaptcha_qsencode ($data) {
  * @param string $path
  * @param array $data
  * @param int port
- * @return array response
+ * @return string response
  */
 function _recaptcha_http_post($host, $path, $data, $port = 80) {
```
andrew-cormick-dockery, line 69: get rid of the $port argument

recaptcha/recaptchalib.php
```diff
@@ -62,33 +64,17 @@ function _recaptcha_qsencode ($data) {
...
-        fclose($fs);
-        $response = explode("\r\n\r\n", $response, 2);
-
-        return $response;
+  $client = new Client();
+  try {
+    $response = $client->request('POST', $host . '/' . $path, ['query' => $data]);
+  }
+  catch (Exception $e) {
+    return NULL;
```
andrew-cormick-dockery, line 75: return NULL if it all fails?  Is this what we want?

johntwyman:
Noting that @andrew-cormick-dockery has been able to verify that this PR and [this one](https://github.com/civicrm/civicrm-core/pull/18872) work as expected.

recaptcha/recaptchalib.php
```diff
@@ -162,32 +147,26 @@ function recaptcha_check_answer ($privkey, $remoteip, $challenge, $response, $ex
...
 
-        $response = _recaptcha_http_post (RECAPTCHA_VERIFY_SERVER, "/recaptcha/api/verify",
-                                          array (
-                                                 'privatekey' => $privkey,
-                                                 'remoteip' => $remoteip,
-                                                 'challenge' => $challenge,
-                                                 'response' => $response
-                                                 ) + $extra_params
-                                          );
+        $response = _recaptcha_http_post (RECAPTCHA_VERIFY_SERVER, "/recaptcha/api/siteverify",
```
andrew-cormick-dockery, line 157: I'd change the name of these variables.  You're using $response both for an input parameter as well as the result.

recaptcha/recaptchalib.php
seamuslee001: Done

recaptcha/recaptchalib.php
seamuslee001: I've switched this to an empty string which I think is more consistent

recaptcha/recaptchalib.php
seamuslee001: I have changed this to be validationResponse now

recaptcha/recaptchalib.php
andrew-cormick-dockery: It's important to understand what's happening here.  A failure here would mean somehow being unable to connect to Google to verify the recaptcha.  Returning an empty string would cause the calling code to view it as a failure, therefore the user would be implored to re-enter the captcha.  It could be argued that if something is wrong with calling the verify server, then this isn't something that should bother the user and the transaction should go ahead.  The administrator needs to know though, and should be alerted that recaptcha verification isn't working.

mattwire:
This looks good to me. Merging based on @andrew-cormick-dockery/@johntwyman review and test.

PR #312, opened 2020-11-25
[REF] Remove xssString as it is providing a false sense of security

seamuslee001:
as per https://github.com/civicrm/civicrm-core/pull/19045

ping @eileenmcnaughton @pfigel @totten @colemanw


seamuslee001:
Jenkins re test this please

PR #313, opened 2020-12-11
Throw exception instead of PHP fatal error so we can handle a parse error in CiviCRM

mattwire:
Calling `trigger_error` is problematic because for a parse error it triggers a PHP error which stops execution immediately and doesn't give chance for it to get logged or for it to be handled within CiviCRM.

A particular example of this is parsing messagetemplates that have smarty enabled. If you add some style tags eg.

```
<style type="text/css">body {
        margin: 0;
        padding: 0;
    }
</style>
```
that will trigger a fatal error in the smarty parser unless you wrap it in literal tags first. But you don't know to wrap it in literal tags because the error never makes it into the CiviCRM logs and gets 🐿️ed away by the default PHP error handler which, on a production site, is often configured to not log anything or is not accessible to the site administrator.

### DIFF
--- a/Smarty/Smarty.class.php
+++ b/Smarty/Smarty.class.php
@@ -1097,10 +1097,16 @@ class Smarty
     function trigger_error($error_msg, $error_type = E_USER_WARNING)
     {
         $msg = htmlentities($error_msg);
-        trigger_error("Smarty error: $msg", $error_type);
+
         // Customised for CiviCRM - this allows us to capture the error messages
         $event = new \Civi\Core\Event\SmartyErrorEvent($error_msg, $error_type);
         \Civi::dispatcher()->dispatch('civi.smarty.error', $event);
+        if ($error_type === E_USER_ERROR) {
+          throw new CRM_Core_Exception("Smarty error: $msg");
+        }
+        else {
+          trigger_error("Smarty error: $msg", $error_type);
+        }
     }
 
 


demeritcowboy:
As a practical thing, I can see this makes some sense. And when I `r-run` it does what it says.

As a theoretical thing, letting users edit a live site's "source code" in realtime in a textbox has always been a bit insane. But I know it doesn't work that way for message templates in practice. You can't force them to test it first on a site where error handling is turned on.

So would this now possibly cause some emails that were "working" before to suddenly stop? I say "working" because the emails would have gone out, just e.g. without the styling. Now they won't go out. That's not necessarily bad, just maybe requires some communication since from the user's point of view it will be "something broke after I upgraded".

mattwire:
> I say "working" because the emails would have gone out, just e.g. without the styling

@demeritcowboy But they wouldn't have gone out? Because the trigger_error(xx, E_USER_ERROR) causes PHP to exit silently and stop processing.

demeritcowboy:
It went out when I tested it pre-patch. I put css in the template, and I can see it comes out mangled (anything between the squiggly brackets is missing), but the email goes.

eileenmcnaughton:
@totten any thoughts

seamuslee001:
@demeritcowboy what CMS were you testing on?

eileenmcnaughton:
Actually I've witnessed what Dave says before in live emails - in the case I'm thinking of the email was fine except near the bottom there was an error so anyone who scrolled right to the bottom would have seen some mangled smarty.

mattwire:
@demeritcowboy Did you have `CIVICRM_MAIL_SMARTY` enabled? I came across this issue with scheduled reminders and also loading a messagetemplate via "preview message template" button in https://lab.civicrm.org/extensions/msgtpltools. In both cases it's a hard fail (ie. PHP error) without this patch. Tested on Wordpress and CiviCRM 5.31 / master.

mattwire:
@demeritcowboy I actually have `CIVICRM_MAIL_SMARTY` disabled on the site exhibiting problems. But it's probably enabled when doing scheduled reminders..

mattwire:
Specifically, my template is triggering a syntax error:
```
line 3]: syntax error: unrecognized tag: margin: 0;\r\n
              padding: 0; (Smarty_Compiler.class.php, line 441)
```
via https://github.com/civicrm/civicrm-packages/blob/master/Smarty/Smarty_Compiler.class.php#L441

eileenmcnaughton:
I guess it would be good to have a test on this covering the different settings & exception modes that might be in play. From the docs the exception handler would alter the way it works and different flows might still use the legacy ignoreException handler which would alter the behaviour

mattwire:
> settings & exception modes that might be in play.
I don't think there are any in play at this stage. It either triggers a PHP error or it does not.

demeritcowboy:
Well the site was in Italian so maybe the error will just show up an hour or two late (zing!)

Seriously, I get the same whether I have CIVICRM_MAIL_SMARTY on or not. It's drupal 7 - maybe that makes a difference. There's no extensions on this site.

Let me try some more things.

demeritcowboy:
I keep getting the same result. Scheduled reminder, scheduled mailing, assignee copy of activity... they all get sent with or without the smarty setting, just the style sometimes gets mangled when smarty tries to process it.

eileenmcnaughton:
But the triggered error is then handled by the error handler isn't it? 

https://www.php.net/function.trigger-error

demeritcowboy:
If we're talking about a live site where you have error reporting not set for these, then no, at least not drupal 7. If I turn on error reporting then yes I see the E_USER message on screen for example for the assignee copy and yes I think that's coming from drupal's error_handler.

mattwire:
@demeritcowboy @eileenmcnaughton The difference here seems to be drupal / wordpress. On drupal `trigger_error` does not cause PHP to exit but on wordpress it does - at least in my testing on a few different sites.

Furthermore, you can trigger this easily by specifying an unknown token which smarty then tries to parse.
Eg. this is enough to trigger it (on wordpress PHP exits silently, on drupal it logs and continues as normal):
```
Test1 token: {test1.test}

Contact display name: {contact.display_name}
```

The error that is logged on drupal (and on wordpress after this PR):
```
 [error] Smarty Error: [in string:
Test1 token: {test1.test}

Contact display name: Matthew Wire line 4]: syntax error: unrecognized tag '' (Smarty_Compiler.class.php, line 585)
```

@christianwach Any ideas about the differences in error trapping on wordpress?

mattwire:
Also linking https://lab.civicrm.org/dev/core/-/issues/58 which sounds very much related.

eileenmcnaughton:
@totten this seems to tie in with discussion last week about wordpress being more fatal on some errors. I think the trigger_error in the code here is right but there is a problem 'somewhere' with Wordpresses error handling - although it might make sense to standardise Civis' handling (remove the last 7 ignoreException) first

seamuslee001:
@mattwire I noticed that in wp.load.php in wordpress (which seems to be its bootstrap file) it sets 

`error_reporting( E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_ERROR | E_WARNING | E_PARSE | E_USER_ERROR | E_USER_WARNING | E_RECOVERABLE_ERROR );`

I think Drupal by default resorts to the global php configuration for the error reporting level.

christianwach:
> Any ideas about the differences in error trapping on wordpress?

@mattwire I was going to point you to what @seamuslee001 found.

demeritcowboy:
But since you can have different error_reporting on live and dev sites it sounds to me like wordpress does not have an error_handler at a "top" level, which isn't wrong just different from drupal which does have one. So in drupal these get caught and either ignored (error_reporting off) or logged (error_reporting all on) and then it continues, and on wordpress they exit and go to wherever error_log is set (e.g. httpd log).

In both cases I think the aim of the PR though is to have them show up more visible to the average office staffer, e.g. in the scheduled job log (if something like scheduled reminders), regardless of error_reporting.

So if removing the ignoreExceptions gets it there, then that's another way to do it, just then you still have the perception problem for drupal users that it was "working" before upgrade, and now broke. One way to address this might be in the error message itself, somehow indicating that "you may not have seen this before because the way smarty errors are logged has now changed", which would also be correct for wordpress.

seamuslee001:
Based on @demeritcowboy 's feedback and also my feelings on this PR I am flagging that I consider this merge-ready and if anyone has any objections please advice but otherwise will merged in a couple of days

eileenmcnaughton:
I'd feel a lot happier if @totten weighed in - this feels like it's not getting to the root of the issue & in the future we are going to wonder if we can unravel this. I guess some pretty thorough commentary in the code would address the latter issue

demeritcowboy:
I'm still worried about the (drupal) user perception point of view: "I had a message template that worked before, and now after upgrade it doesn't send emails anymore." Yes they should fix their buggy smarty code, but are we telling them enough to avoid frustration.

Thinking out loud I wonder if it could be handled by a pre-upgrade check, something like run `smarty->fetch('string: ' ...)` on each message template and if it throws an exception warn the user at that time.

totten:
Well that's fun 🙃  It seems like we have a primary question (ie *how to classify a Smarty parse error*), and then we have a bunch of technical happenstance about what happens based on the classification. Let me be a little pedantic in the primary question.

> But they wouldn't have gone out? Because the trigger_error(xx, E_USER_ERROR) causes PHP to exit silently and stop processing.

OK, so, right now, `Smarty.class.php` is written to `php.net` specifications. To understand the current classification (`E_USER_ERROR`), one should go back and re-read the [php.net definitions](https://www.php.net/manual/en/errorfunc.constants.php):

* __`E_USER_ERROR`__:  User-generated error message. This is like an `E_ERROR`, except it is generated in PHP code by using the PHP function trigger_error(). 
* __`E_ERROR`__:  Fatal run-time errors. These indicate errors that can not be recovered from, such as a memory allocation problem. Execution of the script is halted. 

A Smarty parse error is classified as `E_USER_ERROR`, which signifies a non-recoverable problem that halts execution. That makes sense for Traditional Smarty Usage where a developer maintains `*.tpl` files for the application UI (ie *this is a class of developer-screw-ups for which we cannot give a decent fallback - the page-view is ruined*). However, for email templates written by web-based users (*especially ones with GUI-admin-tools and with batched execution via cron-workers*)...  a Smarty parse error should be treated as recoverable. (*For cron-workers, we should log a failure in the immediate task - and persist in running other tasks. For GUI-admin-tools, we should persist in rendering the admin UI -- and encourage the admin to revise the template.*)

If I'm understanding correctly, WordPress's default behavior (*treat `E_USER_ERROR` as non-recoverable/fatal*) complies with the `php.net` definition better than Drupal's default behavior. But the classification as `E_USER_ERROR` is incorrect, and the behavior of Drupal's less-compliant policy is (has-been-to-date) more forgiving/agreeable.

The stated intent in switching from `E_USER_ERROR` to an `Exception` is to make it recoverable. I agree with this. Since the classification in this case is wrong, it's probably more appropriate to fix the classification than to coax WP into a non-compliant behavior. (*Aside: You might alternatively change the classification by demoting it to a different `E_USER_WARNING` or `E_RECOVERABLE_ERROR`  during email evaluation... though IMHO a new exception class `Civi\Core\Exception\SmartyException` would be more understandable to most devs.*)

Question: Are we going to actually sprinkle-in some more `try/catch` blocks to handle this recoverability?

> Thinking out loud I wonder if it could be handled by a pre-upgrade check, something like run smarty->fetch('string: ' ...) on each message template and if it throws an exception warn the user at that time.

+1 for more validation. I wonder if Smarty 2 has APIs to parse/validate without executing. (Actually, if the main concern is specifically with CSS and is dependent on CIVICRM_MAILING_SMARTY and/or the specific template use-case, then it may be better to focus on those angles rather than Smarty validation per se.)

But even if we had a Smarty validator API available, we're unlikely to identify every problem upfront this way. So it's probably still important to be intentional about the try/catch/recover scenarios.

seamuslee001:
So at the moment we have a few different places where templates can be invoked

1. Scheduled Reminders
2. CiviMails
3. Receipt emails
4. Once off email
5. Thank You Emails

In regards to item 1

the SMS version catches a CRM_Core_Exception https://github.com/civicrm/civicrm-core/blob/e276d2327cae113f11483e1e6d7a55dcba87cfbd/CRM/Core/BAO/ActionSchedule.php#L575 and the Mail uses CRM_Utils_Mail::send so will silently catch things and also in the token processing it looks like we may already handle things but not 100% sure https://github.com/civicrm/civicrm-core/blob/e276d2327cae113f11483e1e6d7a55dcba87cfbd/CRM/Core/BAO/ActionSchedule.php#L287

For Item 2 I would think Flexmailer and also Core CIviMail have handled such problems ok for a while if there was an issue surely it would have come through the queue by now

For Items 3,4  I'm guessing the form layer should be able to handle a CRM_Core_Exception. 

for Item 5 I'm guessing we may want some catching here https://github.com/civicrm/civicrm-core/blob/e276d2327cae113f11483e1e6d7a55dcba87cfbd/CRM/Activity/BAO/Activity.php#L1114 but again I would have thought that the form layer should handle Exceptions appropriately

eileenmcnaughton:
My general assumption is that where we see ignoreException in the code it's an indicator of potential issues - this are still a small handle left to work through

Screenshot from 2020-12-22 14-23-27

eileenmcnaughton:
@totten - I'm struggling with the premise that E_USER_ERROR is a fatal error - when I look I see things that suggest it is unless you have an appropriate error handler - eg

https://github.com/Seldaek/monolog/issues/954#issuecomment-299905537

mattwire:
@eileenmcnaughton @seamuslee001 @totten @demeritcowboy Thankyou for your comments and investigation. I'm not quite sure what is the best way to proceed here.

Depending on your PHP settings you currently get (if you have a smarty "syntax" error):
1. By default in Wordpress (or if E_USER_ERROR is configured as a PHP fatal error): Scheduled reminders (and other messages?) don't go out and you get no notification/indication as to why.
2. By default in Drupal messages do go out but have (maybe) not been parsed correctly and will end up being delivered corrupted.

To me it makes more sense that:
1. The admin is notified in some way.
2. Potentially corrupted messages do not get sent out.

If we replace the `trigger_error` entirely with a throw `CRM_Core_Exception` then you'll automatically achieve 1 and 2. The *issue* is that you may not get notified in an obvious/visible way but you *will* find something in the logs / scheduled job log. It seems that would be better than the current situation and we could then work further on improving catching errors / notifying if required.

eileenmcnaughton:
@mattwire can't we configure an error handler to catch trigger_error? That feels more 'correct' than hacking smarty if it works - ie per that link above.

But to the larger issue - I think that if we can figure out 

1. The admin is notified in some way.

Then it seems pretty clear that 2 makes sense. 

I have seen cases where messages have had corruption at the end of the message & it has not been picked up long term because it was basically pretty subtle but notifying to fix is better. If feels like to do that we want something where an admin email gets a mail

twjordan:
Just ran into this and knowing the error easier would have saved a lot of time!

mattwire:
@eileenmcnaughton @demeritcowboy where are we with this?

> 1. The admin is notified in some way.

Currently there is *nothing*. With this patch we get an exception which makes it's way into the log files. There are many situations where things go wrong and stuff is either recorded in the scheduled job logs or the civi logs but there is no explicit "notification". So I would say that this shouldn't hold up this being merged if we're happy otherwise.  What do you think?

As far as I can tell, anything else requires the sysadmin to make configuration changes to the default wordpress install which isn't likely to happen in most cases I would think.

demeritcowboy:
My vote is +1 on the concept. Reading thru the thread again there is maybe some holdup on agreement as to whether this should be a "recoverable error" or a hard fail, i.e. whether there's enough try/catching/error-handling farther down the line. As a specific example if you have two scheduled mailings and the first has css without `{literal}`, it will now prevent the second mailing from going too. On wordpress that's maybe no different from before, but that is a change on drupal, and doesn't match the concept of "recoverable error".

Do any extra additions need to hold up this PR - no, but I'm assuming there is agreement that this should be a "recoverable" error as opposed to a hard fail. If not then that might be partly where we're stuck.

eileenmcnaughton:
@demeritcowboy I guess my question is why trigger_error doesn't get handled by the error handler that CiviCRM registers (useException) - ie per the comment linked to here https://github.com/civicrm/civicrm-packages/pull/313#issuecomment-749286166

demeritcowboy:
Is useException just for PEAR errors? I suppose I should look at what it actually does.

eileenmcnaughton:
@demeritcowboy perhaps it IS just PEAR errors & the trick would be to make it catch things like this

seamuslee001:
@demeritcowboy @eileenmcnaughton correct it is only about PEAR_Errors in useException, trigger_error is delegated to PHP/CMS to handle

So I think that now that our default PEAR_Error handler is useException I think I am more comfortable with this. The only issue will be is for system jobs and making sure that an Exception here doesn't cause all jobs to stop.

eileenmcnaughton:
@seamuslee001 so I guess my issue is still that I think we should figure out how to make our error handler handle errors rather than hack a package to throw a proprietary exception

seamuslee001:
So are you thinking we need to do something like this? https://www.php.net/manual/en/function.set-error-handler.php but I guess for those CMSes already doing their thing like Drupal here https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Core/DrupalKernel.php#L1044 would we want to override that with our own thing?

demeritcowboy:
@eileenmcnaughton A global set_error_handler I don't think would be desired because back to what this issue was originally about, if you have error_reporting turned down like on a live site, I wouldn't want civi to be overriding what I've set for all errors. The underlying problem as I understand it is "when used for mailings this error that smarty is throwing is not really an E_USER_ERROR".

seamuslee001:
More so in my understanding that when scheduled jobs are executed on systems such as WordPress any E_USER_ERRORs such as this will cause all scheduled jobs to stop processing because the error_reporting level / lack of any set_error_handler being set causes them to be treated as a PHP Fatal error

demeritcowboy:
I suppose if civi had a global error_handler that still respected error_reporting, then that might work.

```
if (error_reporting() & $errno) {
  // do civi logging
}
return FALSE; // don't override what happens above us
```

eileenmcnaughton:
So we can set error handlers in the way that temporary error scopes are done too - ie they are reversed when the object is deconstructed - so we could do that before we render a message

demeritcowboy:
Yes I think so. Although it would then log all errors within that scope regardless of error_reporting, but maybe ok if the scope is limited.

demeritcowboy:
I just did a POC. There's (at least) two options:
  * Like drupal currently, where it logs it but then still sends it.
  * Log it, but don't send it. This kind of makes more sense to me. That's what this patch does:

```diff
diff --git a/CRM/Mailing/BAO/Mailing.php b/CRM/Mailing/BAO/Mailing.php
index 269bf3d8ea..bf1d734a6b 100644
--- a/CRM/Mailing/BAO/Mailing.php
+++ b/CRM/Mailing/BAO/Mailing.php
@@ -1177,7 +1177,17 @@ ORDER BY   civicrm_email.is_bulkmail DESC
     ) {
       $htmlBody = implode('', $html);
       if ($useSmarty) {
-        $htmlBody = $smarty->fetch("string:$htmlBody");
+        // This would be more generic - proof of concept here.
+        set_error_handler(['CRM_Mailing_BAO_Mailing', 'handleError']);
+        try {
+          $htmlBody = $smarty->fetch("string:$htmlBody");
+        }
+        catch (CRM_Core_Exception $e) {
+          restore_error_handler();
+          // it was logged in the handler already, so just abort sending
+          return NULL;
+        }
+        restore_error_handler();
       }
       $mailParams['html'] = $htmlBody;
     }
@@ -3117,4 +3127,11 @@ ORDER BY civicrm_mailing.name";
     return $r;
   }

+  public static function handleError($errno, $errstr, $errfile, $errline) {
+    if (error_reporting() & $errno) {
+      \Civi::log()->error($errstr); // or whatever
+    }
+    throw new CRM_Core_Exception('bad mail');
+  }
+
 }
```

eileenmcnaughton:
@demeritcowboy that looks generally good to me - I've asked @totten to comment. My thoughts are mostly around
1) where should the error handler sit - perhaps on the temporaryErrorScope class
2) how generic it should be - this is pretty tightly tailored to mail - but only because it throws 'bad mail' not $errno & $errstring
3) how high up in the class it would sit - I feel like ideal would be to be have the civi error handler in play when CiviCRM is 'in charge' & then hand back to the CMS level when Civi hands back the reins. That's a bigger change & quite reasonably out of scope but it could be we instantiate a temporary error scope at the start of the function which naturally reverts when the function ends and it is deconstructed

mattwire:
@demeritcowboy The reason I originally targetted the smarty package directly is because we call `$smarty->fetch()` from all over the place. There are at least two places in BAO/Mailing.php and if we're using flexmailer it's triggered from `TokenCompatSubscriber::onRender()`.
So we really don't want error handling at the `$smarty->fetch()` call. *Possibly* it would work if we implemented error-handling in the `CRM_Core_Smarty::fetch()` function directly?

I don't think we should be sending out any email that didn't parse properly. It risks at best looking unprofessional and at worst exposing something we didn't want to - eg. what if there was something that could be exploited in an if statement that didn't get "parsed" out or something. The fact that broken emails have been going out on drupal before doesn't mean they should keep doing so. My main concern is that errors get logged so that someone investigating can *see* that something went wrong and do something about it.

mattwire:
@eileenmcnaughton @demeritcowboy @totten I feel like we're going around in circles with this a little bit :-)

We've worked out that:
1. Smarty "fatal" errors are not reported on Wordpress and emails do not go out.
2. Smarty "fatal" errors are not fatal on drupal and broken emails do go out.
3. Smarty errors should not trigger a PHP fatal error because then CiviCRM can't handle it.

We already modify the Smarty class and the specific function that I'm targetting in this PR to add a Smarty error event that is designed to capture smarty errors and be loggable. But turns out that doesn't work on Wordpress by default because of the the fatalness of E_USER_ERROR so we need to improve that.

And now we're looking at adding in custom error handlers while at the same time we've been trying to remove use of them in core (the ignoreException vs useException stuff). We don't need the custom error handlers at all if we just patch at the source of the error and we are sure that no matter what route we run smarty we'll get an error reported if something is not right.

I'm more than happy to add in an pre-upgrade message into CiviCRM core along the lines of *Smarty error reporting has been improved. If you have any messagetemplates that are not formatted correctly for Smarty this may result in errors being logged and emails not being sent out.*

demeritcowboy:
I'm fine with the idea of changing what smarty is doing in this situation. One thing I didn't notice before though - it looks like _sometimes_ if you make a smarty error in a regular tpl file, like adding some `<style>`'s without literal, then oddly before the patch it shows/logs the error, but with the patch it completely disappears! I'm going to try to trace that and see what's happening and how widespread that is.

demeritcowboy:
Ok I'm not sure what that was. Once I turned on xdebug it made it work as expected, and then it stayed working as expected when I turned it off again. Sorry for the false alarm.

eileenmcnaughton:
Just noting that I did a patch to handle smarty errors in the tokenCompatSubscriber in https://github.com/civicrm/civicrm-core/pull/19862 just to test to https://github.com/civicrm/civicrm-core/pull/18466

It worked but was just a rough cut in order to test the other PR. However, it at least confirmed to me that a non-packages hack was straight forward as an approach.

(note from a merge point of view I'm gonna leave this to @totten as I'm not comfortable enough with this approach to merge it myself)

mattwire:
@eileenmcnaughton The problem with https://github.com/civicrm/civicrm-core/pull/19862 is it only targets the use of smarty in one place. Everywhere that smarty is invoked would need to be patched in a similar way which doesn't seem very reliable/maintainable compared with "fixing" in a single place at source per this PR.

mattwire:
@eileenmcnaughton @demeritcowboy Adding comments from https://github.com/civicrm/civicrm-core/pull/21122 for reference. I'm more than happy to be persuaded otherwise if my assertions are incorrect?

@demeritcowboy There are two issues:
1. Why can we not "hack" the smarty package now?, when we've already done so a year ago on the same line: https://github.com/civicrm/civicrm-packages/pull/292/files
2. As far as I can see *any* solution that is not in the smarty (civicrm-packages) package will require implementation in multiple places in core and extensions:
  - Example where an extension would need to implement the fix in civicrm-packages#313: https://lab.civicrm.org/extensions/mosaico_smarty/-/blob/main/Civi/Token/SmartyTokenSubscriber.php#L32
  - There are multiple examples in core/extensions where a smarty tpl is retrieved and rendered using CRM_Core_Smarty. If any of these templates had an unescaped (ie. no literals) html style block then Wordpress (or any system configured with the standard wordpress php error levels) will fail silently. My testing at the time and my understanding is that every single place this happens would need to implement a "fix" for smarty unless we do it in the upstream package.. where we've already made modifications..

demeritcowboy:
I don't have an answer for item 1 but for 2 in pseudocode I was thinking something like this, which should be an almost equivalent (global) replacement, but is 100% theory at the moment:

```php
function myextension_civicrm_config() {
  set_error_handler('myextension_something');
}

function myextension_something($etype, $emsg, ...) {
  if ($etype == E_USER_ERROR && $emsg starts with "Smarty error: ") {
    throw new CRM_Core_Exception($emsg);
  }
}
```

demeritcowboy:
@mattwire This seems to do it: https://lab.civicrm.org/extensions/smarter

eileenmcnaughton:
Generally we try, over time, to remove hacks where we can from packages & not add more - the one we added seems like it could also have been done outside of smarty. 

In general this isn't an issue that has had a huge amount of reporting over the years so fixing in an exception does seem OK if this is more of a word press thing. As you know I think it makes sense to fix for message templates & for extensions to handle themselves as they require.

But, I'll leave this one to others.

eileenmcnaughton:
Now that all user string smarty parsing in core goes through one function I've put up a PR to fix this old issue in core

https://github.com/civicrm/civicrm-core/pull/25334

I'm hoping we can also remove the exisitng hack (I'm looking into all our smarty hacks right now)

demeritcowboy:
While it doesn't fully address everything here, this should address some of it: https://github.com/civicrm/civicrm-core/pull/25334

demeritcowboy:
I guess I should read the previous comment. 'Twas already noted, but now merged.

eileenmcnaughton:
This is kinda irrelevant now that we are getting off Smarty2

PR #315, opened 2021-02-06
dev/core#2366 - php 7.4 - Deprecated function get_magic_quotes_gpc

demeritcowboy:
https://lab.civicrm.org/dev/core/-/issues/2366

* The first `if` block never runs because get_magic_quotes_gpc() always returns false.
* The second `if` block always runs because php version is always > 5.3
* So just do the second block.

### DIFF
--- a/IDS/Monitor.php
+++ b/IDS/Monitor.php
@@ -333,16 +333,8 @@ class IDS_Monitor
             }
         }
 
-        // check for magic quotes and remove them if necessary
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Unit test is at https://github.com/civicrm/civicrm-core/pull/19851

seamuslee001:
Looks good merging now

demeritcowboy:
Thanks!

PR #316, opened 2021-02-20
php 7.4 deprecated squiggly brackets - html/quickform date

demeritcowboy:
Easiest way to trigger is just visit civicrm/admin/setting/date?reset=1


demeritcowboy:
Thanks!

PR #322, opened 2021-04-28
Fix brackets in DataObject

seamuslee001:
This fixes some brackets because in https://github.com/civicrm/civicrm-packages/pull/320 it added an extra set of `()` around the 2nd set of options which wasn't really needed here


eileenmcnaughton:
yep - looks good

PR #323, opened 2021-05-29
[php8-compat] Fix calling method_exist with paremeter that is bool not an object in…

seamuslee001:
… php8

like https://github.com/civicrm/civicrm-core/pull/20443 this just adds a check to make sure that $result is actually a class object


eileenmcnaughton:
what else would it be? is this a flaw in the upstream code or in our usage/ hacks to this?

seamuslee001:
@eileenmcnaughton it was complaining it was an integer iirc so maybe but it was pretty early on in the process so this resolves it at least

seamuslee001:
test this please

eileenmcnaughton:
Ok - I guess this particular function is already hacked so it might be our hack that is bad. I'm not keen on hacking packages but this just an extra check I guess

PR #324, opened 2021-05-29
DB/DataObject: use mb_strtolower because in turkish, INSERT becomes Insert

mlutfy:
To reproduce, make sure you have the Turkish locale enabled (`dpkg-reconfigure locales` and enable `tr_TR.utf8`), then you can test with:

```
$ php -r 'setlocale(LC_ALL, "tr_TR.utf8"); echo strtolower("INSERT") . "\n";'
Insert
$ php -r 'setlocale(LC_ALL, "tr_TR.utf8"); echo mb_strtolower("INSERT") . "\n";'
insert
```

Huge thanks to @demeritcowboy for the lead.

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -2717,7 +2717,8 @@ class DB_DataObject extends DB_DataObject_Overload
             continue;
           }
             } else {
-                switch (strtolower(substr(trim($string),0,6))) {
+                // civicrm-packages#324 Use mb function because if setlocale is set to tr_TR.utf8, INSERT would become Insert
+                switch (mb_strtolower(substr(trim($string),0,6))) {
 
                     case 'insert':
                     case 'update':
@@ -2754,8 +2755,9 @@ class DB_DataObject extends DB_DataObject_Overload
 
         // CRM-18093 starts.
         // CRM-20445 starts Strip any prepended comments
+        // civicrm-packages#324 Use mb function because if setlocale is set to tr_TR.utf8, INSERT would become Insert
         $queryString = (substr($string, 0, 2) === '/*') ? substr($string, strpos($string, '*/') + 2) : $string;
-        $action = strtolower(substr(trim($queryString),0,6));
+        $action = mb_strtolower(substr(trim($queryString),0,6));
         // CRM-20445 ends
 
         if (!empty($_DB_DATAOBJECT['CONFIG']['debug']) || (defined('CIVICRM_DEBUG_LOG_QUERY') && CIVICRM_DEBUG_LOG_QUERY)) {


mlutfy:
Any thoughts how we could test this? Test nodes do not have locales such as tr_TR.utf8, but we could add them..

demeritcowboy:
Oh wow so it _was_ turkish. For testing do you mean a unit test? If the nodes have the non-utf8 tr_TR you should still get a fail calling that function if you set that locale.

mlutfy:
yep unit tests. I will try to install the locales on the test nodes, and add a test.

totten:
Regarding test nodes -- I tried this step on a couple nodes:

```
$ php -r 'setlocale(LC_ALL, "tr_TR.utf8"); echo strtolower("INSERT") . "\n";'
Insert
$ php -r 'setlocale(LC_ALL, "tr_TR.utf8"); echo mb_strtolower("INSERT") . "\n";'
insert
```

It produced the same quirky result for `strtolower()` (and valid result for `mb_strtolower()`). So I don't think you need to install anything more to make the test-server a comparable runtime.

Tangents:

* If one did need to dig into the locale-installation there, it might help to know that min/dfl/max/edge profiles each have their own locale folder. This is automatically wired-up with an env-var `LOCALE_ARCHIVE` =/nix/var/nix/profiles/bknix-{$PROFILE}/lib/locale/locale-archive`.

*  On my local macOS laptop, both commands yield the expected `insert`. Not certain which reason applies there, but it probably doesn't matter too much -- we don't need to actively reproduce the old bug everywhere. All that really matters is that `mb_strtolower()` also works here.)

demeritcowboy:
@totten On your mac, what is the return value of setlocale. My guess is FALSE, meaning it didn't actually set it.

totten:
@demeritcowboy Quite right -- on macOS, `setlocale(...tr_TR.utf8...)` returns `FALSE`; on Linux, it returns the name of the locale.

nixpkgs on linux appears to include several hundred locales - based on GNU's stack (libc+locales).

```
$ localedef --list-archive /nix/var/nix/profiles/bknix-min/lib/locale/locale-archive  | wc
    855     855    8855
$ localedef --list-archive /nix/var/nix/profiles/bknix-min/lib/locale/locale-archive  | grep utf8 | wc
    311     311    3568
```

I'm not certain how nixpkgs wires-up the locales on darwin, but I think all the localization data is implemented separately. (Theory: The choice of cc+libc+locales is aligned with macOS conventions.) Interestingly, Darwin-Turkish and GNU-Turkish seem a little different - ie `tr_TR.utf8` vs `tr_TR.UTF-8`. The behavior of `strtolower()` in Darwin-Turkish satisfies my English-speaker intuition - ie `strtolower('INSERT') === 'insert'`.

```bash
$ uname
Darwin
$ php -r 'var_export(setlocale(LC_ALL, "tr_TR.utf8")); echo "\n"; echo strtolower("INSERT") . "\n";'
false
insert
$ php -r 'var_export(setlocale(LC_ALL, "tr_TR.UTF-8")); echo "\n"; echo strtolower("INSERT") . "\n";'
'tr_TR.UTF-8'
insert
```

demeritcowboy:
Failing unit test: https://github.com/civicrm/civicrm-core/pull/20557

I think this is mergeable after.

PR #325, opened 2021-07-13
upper modifier with accents / umlaut

Anatoleallain:


demeritcowboy:
I know this comes from https://civicrm.stackexchange.com/questions/39967/upper-case-in-name-with-accents and there I suggested hacking this file as a quick local fix, but looking more there's precedent for adding one to civi core instead of changing a 3rd party package. See https://github.com/civicrm/civicrm-core/blob/b6601e07a902dcffcb9c46e113c52d7020aa8066/CRM/Core/Smarty/plugins/modifier.mb_truncate.php which is the same thing but for truncate instead of upper. So I would suggest making a similar file there called modifier.mb_upper.php instead.

Note that in modifier.mb_truncate.php it does some checking if the mb_ function exists, but you no longer have to do that since it's [required for civi](https://docs.civicrm.org/installation/en/latest/general/requirements/#required-for-civicrm-core) now. So it would be a very simple file.

demeritcowboy:
Closing - see https://github.com/civicrm/civicrm-core/pull/20892

PR #319, opened 2021-04-24
dev/core#2558 Insert blank MYSQLTIMESTAMP fields as NULL rather than 0

ufundo:
Extends the behaviour for date fields to timestamp fields also. This was hinted at in the comment but maybe there was a reason not to?

Seems to fix https://lab.civicrm.org/dev/core/-/issues/2558 for me, but is way out of my comfort zone!

PR #326, opened 2021-07-17
Also check timestamp types for update() when trying to update a timestamp field with a blank value

demeritcowboy:
Overview
In https://github.com/civicrm/civicrm-packages/pull/319 insert() was patched to also look for timestamp db fields when inserting a blank date in a field that allows nulls. But update() was not also patched so it has the same problem with timestamp types.

Before
See unit test at https://github.com/civicrm/civicrm-core/pull/20891 - it fails with:
DB Error: unknown error
Incorrect datetime value: '0' for column `whatever` at row 1

After
stores null in the db

Technical Details
It only fails if sql_mode has NO_ZERO_DATE.

Comments
https://github.com/civicrm/civicrm-core/pull/20476

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -1472,7 +1472,7 @@ class DB_DataObject extends DB_DataObject_Overload
             // DATE is empty... on a col. that can be null..
             // note: this may be usefull for time as well..
             if (!$this->$k &&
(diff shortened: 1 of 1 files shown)


seamuslee001:
Looks fine to me

PR #327, opened 2021-08-03
Remove illegal character notice when downloading a vcard with a special character

kainuk:
## How to reproduce

* Configure your CiviCRM, so that notices are shown.
* Select a contact with a special character in the name.  In the standard test set _Allan Müller_ is a good candidate.
* Select the action (1) vcard (2)
vcard
* Now a vcard is downloaded on your local disk.
* Refresh your screen

## Before
A notice is shown
notice2

## After
The notice is not there

## Technical details

The `//IGNORE` option is documented at https://www.php.net/manual/en/function.iconv.php . Instead of throwing a notice, it just removes the illegal character. This makes the line shorter so the the test on length (after the change) is still valid.

### DIFF
--- a/Contact/Vcard/Build.php
+++ b/Contact/Vcard/Build.php
@@ -2063,7 +2063,7 @@ class Contact_Vcard_Build extends PEAR
                     $strlen = mb_strlen($line, 'UTF-8');
                 }
                 foreach ($charsets as $charset) {
-                    $iconvd = iconv('UTF-8', $charset, $line);
+                    $iconvd = iconv('UTF-8', $charset.'//IGNORE', $line);
                     if (strlen($iconvd) == $strlen) {
                         $lineCharset = $charset;
                         $line = $iconvd;


demeritcowboy:
The idea sounds ok but when I try it on the test site here I get a new error, although the downloaded file is ok and the same as before the patch:

`Notice: iconv(): Unknown error (84) in Contact_Vcard_Build->charsetRecode() (line 2066 of /home/jenkins/bknix-dfl/build/pkg-327-6ajok/web/sites/all/modules/civicrm/packages/Contact/Vcard/Build.php)`

Maybe this new error depends on the version of iconv or what other language-related packages are installed on the server.

On the bigger picture, I wonder if this function is even still needed - are there still people using versions of outlook that can't import utf-8? This function doesn't change the string itself, it just tries to determine what to put for CHARSET on the line beside it, looking for the minimum charset possible. But nowadays could it just always put UTF-8?

jenkins retest this please (I don't think the fails are related)

demeritcowboy:
Actually there are some situations where it might recode the line, but still unless old outlook needs to be supported if it could just simply label it as utf-8 and never recode that sounds nice.

kainuk:
It is a good idea to remove code to make it simpler. The current version does not work well with Google Contacts for the Łąchowski-Yadav family. (It does with Thunderbird).

kainuk:
We tested some vcards with UTF encoding in the latest versions of Microsoft. Both did not work.

With no encoding
```
BEGIN:VCARD
VERSION:2.1
FN:Łąchowski-Yadav family
N:Łąchowski-Yadav family;;;;
ADR;TYPE=HOME,PREF:;;653F Caulder Pl S;Wilkes Barre;Pennsylvania;18761;United States
END:VCARD
```
and UTF-8 notation
```
BEGIN:VCARD
VERSION:2.1
FN;ENCODING=8BIT;CHARSET=UTF-8:Łąchowski-Yadav family
N;ENCODING=8BIT;CHARSET=UTF-8:Łąchowski-Yadav family;;;;
ADR;TYPE=HOME,PREF:;;653F Caulder Pl S;Wilkes Barre;Pennsylvania;18761;United States
END:VCARD
```
Both did not work. So, I will leave it as it is for now.

PR #328, opened 2021-08-11
Replace Drupal 9 user function, function getUsername is no more valid

sunilpawar:
Drupal 9 no more support `getUsername` function, this lead to fatal error while accessing the images through kcfinder.

Supported function in Drupal 8 and Drupal9 is `getAccountName`.


demeritcowboy:
Thanks @sunilpawar

PR #329, opened 2021-08-24
Fixes dev/core#2769 remove quickform hack - we are no longer calling this rule

eileenmcnaughton:
It adds to confusion and we generally try to get our packages back to 'stock'

https://lab.civicrm.org/dev/core/-/issues/2769#note_63305

UPDATE - I thought https://github.com/civicrm/civicrm-core/pull/21169#discussion_r694017240 was merged - needs to be merged first

@seamuslee001 this is a follow up to the changes I did for https://lab.civicrm.org/dev/core/-/issues/2769 - ie we no longer need to hack quickform because we are doing validation within CRM and hacking it has downsides

- confusion - per the gitlab
- makes people think that hacking packages is 'ok' since we did it in place x
- at some level all hacks need to be maintained & re-applied on upgrade - they have an ongoing cost

I don't know what over QF hacks we still have ....

### DIFF
--- a/HTML/QuickForm/Rule/Email.php
+++ b/HTML/QuickForm/Rule/Email.php
@@ -37,28 +37,7 @@ require_once 'HTML/QuickForm/Rule.php';
  */
 class HTML_QuickForm_Rule_Email extends HTML_QuickForm_Rule
 {
-
-    /**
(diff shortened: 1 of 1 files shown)


seamuslee001:
Works for me

PR #331, opened 2021-09-10
Use HTML_Purifier from composer

colemanw:
Use HTML_Purifier from composer instead of the one included with the IDS package


demeritcowboy:
One other thing to note with IDS is it only kicks in for non-admins. It doesn't even instantiate for admins.

colemanw:
Test run is a miserable failure but that's to be expected since the test isn't running with civicrm/civicrm-core#21432

demeritcowboy:
Merge-ready pending the core companion.

PR #332, opened 2021-09-14
Beautifier - avoid e_notice in php 7.4 when acting on first comment token

demeritcowboy:
Running GenCode for e.g. a DAO gives an E_NOTICE when it tries to get the previous token while processing the header comment, since there is no previous token.


seamuslee001:
This seems fine to me

PR #334, opened 2021-09-25
Use HTML_Purifier from composer and rely on autoloader

colemanw:
…


seamuslee001:
Jenkins re test this please

PR #336, opened 2021-12-11
Update jquery.tokeninput.js to 1.6.1

jitendrapurohit:
Update jquery.tokeninput.js from https://github.com/loopj/jquery-tokeninput/releases/tag/v1.7.0


colemanw:
@jitendrapurohit I'm fine with merging this, but this package is no longer used anywhere in core. What prompted you to submit this?

KarinG:
@colemanw - we ran into a bug with the autocomplete existing contact element in webform civicrm module -> which to date has been using this package. If it's no longer used in Core then maybe we should just handle this on our webform civicrm side of things?

colemanw:
@KarinG full circle :) Back in the early days of `webfom-civicrm 2.x` I required the `libraries` module and people had to go and download `tokeninput.js` themselves and add it to their `sites/all/libraries` directory. It was a total PITA so I ended up scrapping that and had `webform_civicrm` just use the copy from CiviCRM - one less install gotcha. Then a few years later on the core-team side of things we ditched `tokeninput.js` in favor of `Select2.js` but for the sake of `webform_civicrm` users I never deleted the package from core.

But I think libraries are easier to manage now in D9. Maybe just require it in the `webform_civcrm` `composer.json`?

colemanw:
This is such a minor version bump, just 1.6.0 -> 1.6.1. Let's just merge it.

KarinG:
🙂 Thank you! And yes I started looking at how eg better_exposed_filters module requires its jquery dependences.

KarinG:
@colemanw would it be possible to port this back to 5.45.*

Because -> Jon's PR -> https://github.com/colemanw/webform_civicrm/pull/680
Fails on 5.45 -> 

image

colemanw:
Sorry there are no more releases planned for the 5.45 series.

KarinG:
🤔 In the past ESR point releases post security updates have included backports of important fixes. I know b/c I've provided/tested some for previous ESRs.

KarinG:
Solved by https://github.com/colemanw/webform_civicrm/pull/729 [essentially removing any dependency on what ships/or not with CiviCRM Core]

PR #335, opened 2021-11-15
[REF] dev/drupal#169 Replace other usages of session_id with CRM_Core…

seamuslee001:
…_Config::singleton()->userSystem->getSessionId

ping @demeritcowboy @MegaphoneJon

PR #338, opened 2022-01-09
dev/drupal#169 - Fix namespacing in kcfinder

demeritcowboy:
Updates https://github.com/civicrm/civicrm-packages/pull/335 to avoid namespace issue.

### DIFF
--- a/kcfinder/core/class/browser.php
+++ b/kcfinder/core/class/browser.php
@@ -526,7 +526,7 @@ class browser extends uploader {
             $this->errorMsg("Unknown error.");
         $filename = basename($dir) . ".zip";
         do {
-            $file = md5(time() . CRM_Core_Config::singleton()->userSystem->getSessionId());
+            $file = md5(time() . \CRM_Core_Config::singleton()->userSystem->getSessionId());
             $file = "{$this->config['uploadDir']}/$file.zip";
         } while (file_exists($file));
         new zipFolder($file, $dir);
@@ -559,7 +559,7 @@ class browser extends uploader {
         }
 
         do {
-            $file = md5(time() . CRM_Core_Config::singleton()->userSystem->getSessionId());
+            $file = md5(time() . \CRM_Core_Config::singleton()->userSystem->getSessionId());
             $file = "{$this->config['uploadDir']}/$file.zip";
         } while (file_exists($file));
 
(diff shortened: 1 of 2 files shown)


tapashdatta:
This works @demeritcowboy. Thanks

colemanw:
Unrelated test fails.

demeritcowboy:
Thanks. I'll put up a 5.45 backport in case there's a .1 release.

totten:
On the explanation, "fix namespacing" is a little opaque. I searched a bit and added a link to Stackexchange, which seems to be the original error report.

So to clarify scope of impact... the original issue `dev/drupal#169` was about environmental-compatibility. Did this regression impact all environments or just some environments? Are we missing some coverage here (either wrt `r-run` or `phpunit`)?

seamuslee001:
@totten it only affected Drupal 8./9 AFAIK as that is the only one where the namespacing is important.

demeritcowboy:
Yes it should just be drupal 8/9.

I'd say the coverage gap on 335 was mostly a review fail, but with a partial excuse. I tested it on drupal 7 and tried to test it on drupal 8 but kcfinder didn't work on drupal 8 so I didn't think too much more about it. Or at least it doesn't work out of the box - obviously some people have a workaround.

And for the description here I should have linked directly to [the comment](https://github.com/civicrm/civicrm-packages/pull/335#commitcomment-63171960) on 335. And yes it's only partly related to dev/drupal169 - 335 was something @seamuslee001 put up as a followup to the fix for dev/drupal169 for completeness.

totten:
Yeah, the disposition of ckeditor/kcfinder/cms-compat is confusing. And testing something like this is... rather persnickity...

FWIW, the report on StackExchange had some D7 indicators (`sites/all/modules` is more common D7), and I was perplexed by the idea that  `civicrm-packages/kcfinder/core/class/browser.php` has different namespacing based on CMS revision. So I retested on D7 - for me, (a) the error arises on D7 and (b) the patch solves it on D7. 👍  

(So in the RN's, I omitted any qualifier about CMS/environment.)

demeritcowboy:
Interesting. I just tried again with my local drupal 7 and it's fine without the backslash. Maybe some other factor? Anyway it seems fixed for everybody now.

Trying to think if I can blame something on composer...

PR #341, opened 2022-04-28
Try and remove usage of deprecated strftime function

seamuslee001:


demeritcowboy:
"Try?"

date() takes a different format so I don't think this will work. The header one is easy but the modifier one might be tough.

seamuslee001:
I'm just seeing how I can try and unbust the edge not actually trying to make it work perfectly but yes this was just a silly attempt at getting a build to go on edge

seamuslee001:
closing in favour of #344

PR #343, opened 2022-05-10
[KCFinder integration] Change working directory before trying to include CiviCRM settings file

jensschuppe:
... as otherwise, this fails for KCFinder scripts being one level deeper in the directory tree (`path/to/doc_root/web/libraries/civicrm/packages/kcfinder`) than `civicrm.config.php` expects (`../../../sites/default`).

Follow-up to #342 with relation to [dev/drupal#179](https://lab.civicrm.org/dev/drupal/-/issues/179), [dev/core#3419](https://lab.civicrm.org/dev/core/-/issues/3419) and [dev/civicrm-asset-plugin#4](https://lab.civicrm.org/dev/civicrm-asset-plugin/-/issues/4)

### DIFF
--- a/kcfinder/integration/civicrm.php
+++ b/kcfinder/integration/civicrm.php
@@ -42,6 +42,7 @@ function checkAuthentication() {
     $current_cwd   = getcwd();
     $civicrm_root  = dirname(dirname(getcwd()));
     $authenticated = false;
+    chdir($civicrm_root);
     require_once "{$civicrm_root}/civicrm.config.php";
     require_once 'CRM/Core/Config.php';
 


demeritcowboy:
jenkins add to whitelist

The way I've seen to handle that type of folder issue for other things is by adding a [settings_location.php](https://docs.civicrm.org/sysadmin/en/latest/troubleshooting/#conf-drupal-sites) file. But if every drupal 9 site would need one for kcfinder then maybe something like this is more appropriate. I have some other thoughts I'll put in the ticket.

ufundo:
superceded by https://github.com/civicrm/civicrm-packages/pull/444 -- kcfinder has been switched to use normal routes so doesn't need specific settings file handling

PR #344, opened 2022-06-21
[REF][PHP8.1] Replace a couple of calls in Smarty code to strftime to…

seamuslee001:
… the equivilant using date to fix deprecation notices


demeritcowboy:
By the way I was looking at other uses of strftime in CRM_Utils_Date and thinking it would make sense to do them like https://github.com/civicrm/civicrm-core/blob/c9a916033dab6fdd50d0047513f2dc6fa44c07ec/CRM/Utils/Date.php#L244. Not sure if you already have those in the queue.

PR #348, opened 2022-07-19
[REF][PHP8.1] Fix some further places where passing in NULL is deprec…

seamuslee001:
…ated in php8.1

ping @eileenmcnaughton @demeritcowboy @totten @colemanw

### DIFF
--- a/IDS/Converter.php
+++ b/IDS/Converter.php
@@ -158,7 +158,7 @@ class IDS_Converter
                 if (preg_match_all('/\d*[+-\/\* ]\d+/', $char, $matches)) {
                     $match = preg_split('/(\W?\d+)/',
                                         (implode('', $matches[0])),
-                                        null,
+                                        -1,
                                         PREG_SPLIT_DELIM_CAPTURE);
 
                     if (array_sum($match) >= 20 && array_sum($match) <= 127) {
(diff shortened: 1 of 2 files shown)


IDS/Converter.php
```diff
@@ -158,7 +158,7 @@ public static function convertFromJSCharcode($value)
                 if (preg_match_all('/\d*[+-\/\* ]\d+/', $char, $matches)) {
                     $match = preg_split('/(\W?\d+)/',
                                         (implode('', $matches[0])),
-                                        null,
+                                        '',
```
demeritcowboy, line 161: 3rd param is an int for preg_split?

demeritcowboy: Otherwise PR looks good.

seamuslee001: ah good point will update to -1

seamuslee001: ok fixed now @demeritcowboy

demeritcowboy: Cool. It turns out -1 and 0 mean the same thing in that spot (previously null would have been cast to 0).

demeritcowboy:
Jenkins retest this please.

(Fails have been seen before I think as intermittent time-related, but let's check.)

PR #353, opened 2022-10-28
DB_DataObject - Fix pervasive warnings on PHP 8.2

totten:
See: https://wiki.php.net/rfc/deprecate_dynamic_properties


demeritcowboy:
Yeah that's gonna be a fun one. I saw that in https://lab.civicrm.org/dev/core/-/issues/3833 and was hoping it would go away if we just ignored it long enough.

PR #354, opened 2022-11-01
[REF][PHP8.2] dev/core#3958 Remove refence to num_rows as this has be…

seamuslee001:
…en removed from pear/db

as per discussion https://lab.civicrm.org/dev/core/-/issues/3958#note_84022 ping @demeritcowboy @totten @eileenmcnaughton


demeritcowboy:
I clicked around a bit. I figure something would fail pretty quickly if it was a problem.

PR #356, opened 2022-12-22
fixes core#4047 - htmlspecialchars called on empty array

MegaphoneJon:
Full details on https://lab.civicrm.org/dev/core/-/issues/4047, but basically - we are sometimes passing empty arrays to this function.  Seems like the other instances that were changed on #346 will always be strings, unlike this one.

### DIFF
--- a/HTML/Common.php
+++ b/HTML/Common.php
@@ -141,7 +141,7 @@ class HTML_Common
         if (is_array($attributes)) {
             $charset = HTML_Common::charset();
             foreach ($attributes as $key => $value) {
-                $strAttr .= ' ' . $key . '="' . htmlspecialchars(($value ?? ''), ENT_COMPAT, $charset) . '"';
+                $strAttr .= ' ' . $key . '="' . htmlspecialchars((is_array($value) ? '' : ($value ?? '')), ENT_COMPAT, $charset) . '"';
             }
         }
         return $strAttr;


HTML/Common.php
```diff
@@ -141,7 +141,7 @@ function _getAttrString($attributes)
         if (is_array($attributes)) {
             $charset = HTML_Common::charset();
             foreach ($attributes as $key => $value) {
-                $strAttr .= ' ' . $key . '="' . htmlspecialchars(($value ?? ''), ENT_COMPAT, $charset) . '"';
+                $strAttr .= ' ' . $key . '="' . htmlspecialchars(($value ?: ''), ENT_COMPAT, $charset) . '"';
```
demeritcowboy, line 144: Is 0 a possible legit value here?

MegaphoneJon: Hmm - good point.  I can't say definitively but I'll account for that.

HTML/Common.php
```diff
@@ -141,7 +141,7 @@ function _getAttrString($attributes)
         if (is_array($attributes)) {
             $charset = HTML_Common::charset();
             foreach ($attributes as $key => $value) {
-                $strAttr .= ' ' . $key . '="' . htmlspecialchars(($value ?? ''), ENT_COMPAT, $charset) . '"';
+                $strAttr .= ' ' . $key . '="' . htmlspecialchars((is_array($value) ? '' : $value ?? ''), ENT_COMPAT, $charset) . '"';
```
seamuslee001, line 144: @MegaphoneJon just for the sake of clarity are you able to put `(` and `)` around the $value ?? '' in the false part of the first ternary operator

MegaphoneJon: Sorry @seamuslee001, just saw this.

seamuslee001:
thanks @MegaphoneJon

PR #357, opened 2022-12-31
Fix broken call to registerElementType in advmultiselect.php

braders:
I've been testing some static analysis tools on the CiviCRM codebase and getting an error that `HTML_QuickForm::registerElementType` is being called statically, despite not being a static method. 

I'm not able to replicate the error whilst using CiviCRM directly, but the error is legitimate. I guess somewhere in Civi the error must be being suppressed...

A bit of digging shows that:

1. `registerElementType` is not actually called outside of `advmultiselect.php` (none of the other field types have a call to `registerElementType`.
2. All that `registerElementType` is adds the three arguments to `$GLOBALS['HTML_QUICKFORM_ELEMENT_TYPES']`. However, $GLOBALS['HTML_QUICKFORM_ELEMENT_TYPES'] is configured with  `advmultiselect` at the top of `QuickForm.php`. Therefore there is no need for `registerElementType` to be called.

Possibly `registerElementType` should actually be made static - I'm not sure if it's likely that third-party extensions are registering their own element types??

As an aside, this is my first contribution to `civicrm-packages`. Am I right in thinking we don't bother trying to raise issues like this with any upstream quickform repositories anymore?

### DIFF
--- a/HTML/QuickForm/advmultiselect.php
+++ b/HTML/QuickForm/advmultiselect.php
@@ -1169,9 +1169,4 @@ class HTML_QuickForm_advmultiselect extends HTML_QuickForm_select
         return $options;
     }
 }
-
-if (class_exists('HTML_QuickForm')) {
-    HTML_QuickForm::registerElementType('advmultiselect',
-        'HTML/QuickForm/advmultiselect.php', 'HTML_QuickForm_advmultiselect');
-}
 ?>


demeritcowboy:
jenkins add to whitelist

demeritcowboy:
This is interesting. What's happening is that at the time that statement executes it has been `include`'d as part of the instance of whatever form class, so is actually running in object context, and it's smart enough to know that. See https://github.com/civicrm/civicrm-packages/blob/01f381d6e68d26616d50dcf8f08dcad223907169/HTML/QuickForm.php#L602. You can see this by adding a call like `\CRM_Core_Error::debug_log_message(print_r($this, true));` just before the call to `HTML_QuickForm::registerElementType` here, and then e.g. visit `civicrm/admin/setting/localization?reset=1` and look in ConfigAndLog.

You can also see it like so:

File a.php:
```php
<?php
class A {
  function aaa() {
    print_r($this);
  }
}
```

File b.php:
```php
<?php
include_once 'a.php';
class B extends A {
  public function bbb() {
    A::aaa();
  }
}
```
Then run `php -r 'require_once "b.php"; $b = new B(); $b->bbb();'`

seamuslee001:
Also just noting the include of advmultiselect in HTML/Quickform.php seems to be a CiviCRM customisation as this is how it ships from upstream https://github.com/civicrm/civicrm-packages/blob/master/_ORIGINAL_/HTML/QuickForm.php#L69

PR #358, opened 2023-01-07
Update class_image.php for PHP8

Michael-Labriola:
"each" removed in PHP8

### DIFF
--- a/kcfinder/lib/class_image.php
+++ b/kcfinder/lib/class_image.php
@@ -116,8 +116,8 @@ abstract class image {
             $img = $image->image;
 
         } elseif (is_array($image)) {
-            list($key, $width) = each($image);
-            list($key, $height) = each($image);
+            $width = $image[0];
+            $height = $image[1];
             $img = $this->getBlankImage($width, $height);
 
         } else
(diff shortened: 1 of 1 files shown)


demeritcowboy:
jenkins test this please

kcfinder/lib/class_image.php
```diff
@@ -116,8 +116,8 @@ final protected function buildImage($image) {
             $img = $image->image;
 
         } elseif (is_array($image)) {
-            list($key, $width) = each($image);
-            list($key, $height) = each($image);
+            $width = $image[0];
+            $height = $image[1];
```
demeritcowboy, line 120: Technically this would fail if somebody did something like `$gd = new image_gd(['w' => 200, 'h' => 200]);`. I don't see it used that way anywhere, but it might be safer to use
```php
$width = array_shift($image);
$height = array_shift($image);
```

Michael-Labriola: That looks good to me.

demeritcowboy: Great. I see this is your first contribution. Thanks for this. For release notes credit can do you do another PR to add yourself to this file: https://github.com/civicrm/civicrm-core/blob/master/contributor-key.yml

demeritcowboy: jenkins add to whitelist

demeritcowboy:
jenkins add to whitelist

PR #360, opened 2023-02-23
(REF) Smarty - Fix PHP 8.1 warning in 'modifier.escape.php'

totten:
Steps to reproduce
------------------

* Use PHP 8.1 and D7 with a default logging policy (eg "All Messages" or "Errors and Warnings")
* Use "Search > Find Contacts"
* On the listing page, observe a bunch of warnings

Before
------

~15 warnings, including ~11 related to `modifier.escape.php`

After
-----

~4 warnings, none involving `modifier.escape.php`

Comment
----------

The numbers

### DIFF
--- a/Smarty/plugins/modifier.escape.php
+++ b/Smarty/plugins/modifier.escape.php
@@ -23,7 +23,7 @@ function smarty_modifier_escape($string, $esc_type = 'html', $char_set = 'ISO-88
 {
     switch ($esc_type) {
         case 'html':
-            return htmlspecialchars($string, ENT_QUOTES, $char_set);
+            return htmlspecialchars($string ?: '', ENT_QUOTES, $char_set);
 
         case 'htmlall':
             return htmlentities($string, ENT_QUOTES, $char_set);


Smarty/plugins/modifier.escape.php
```diff
@@ -23,7 +23,7 @@ function smarty_modifier_escape($string, $esc_type = 'html', $char_set = 'ISO-88
 {
     switch ($esc_type) {
         case 'html':
-            return htmlspecialchars($string, ENT_QUOTES, $char_set);
+            return htmlspecialchars($string ?: '', ENT_QUOTES, $char_set);
```
demeritcowboy, line 26: Mixed thoughts:
* Upstream does not address this and lets it error the same as now.
* There's the time savings from doing it here, but possibly hides some code bug. (Although of all the stricter php warnings I think this "null should be a string" one is the least useful one.)
* If we are going to do it here, then should the whole function just return early with `''` if $string is null?

demeritcowboy: So this issue is kind of making civi unusable if you have your settings turned up to 11. I think I'd be ok to merge a version of it that is like the third bullet above. And while it's not in v3 of smarty, looking at the latest master it does this: `$string = (string)$string;`, so we could do the same:

```diff
 */
function smarty_modifier_escape($string, $esc_type = 'html', $char_set = 'ISO-8859-1')
{
+  $string = (string)$string;
    switch ($esc_type) {
        case 'html':
            return htmlspecialchars($string, ENT_QUOTES, $char_set);
```

although this seems more efficient when it's null:

```diff
 */
function smarty_modifier_escape($string, $esc_type = 'html', $char_set = 'ISO-8859-1')
{
+  if ($string === NULL) {
+    return '';
+  }
    switch ($esc_type) {
        case 'html':
            return htmlspecialchars($string, ENT_QUOTES, $char_set);
```

colemanw:
@totten ^

demeritcowboy:
https://github.com/civicrm/civicrm-packages/pull/366 merged

PR #361, opened 2023-05-01
[REF] Fix issue where by using the i18n translation form with RichTex…

seamuslee001:
…tEditor widgets was causing data to be escpaed in the database

@demeritcowboy I believe this resolves your testing issue found here https://github.com/civicrm/civicrm-core/pull/26135

The problem being is that the data is being escaped as part of exportValues before being sent into the database because by default the widgets have a collapsed class added here https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/I18n/Form.php#L79 as well as the wysiwyg class here https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Form.php#L476 which means that this if statement returns false for them

ping @mlutfy


demeritcowboy:
The other PR was merged. Not sure what you want to do with this one. The fails are real but seem like just some undefined variables that would need fixing in core. The upgrade fails are maybe something about the test config in this repo - like it needs updating after the recent core PR that upped the minimum version?

PR #363, opened 2023-06-10
[REF][PHP8.2] IDS fixes for dynamic property deprecation

braders:
In PHP 8.2 dynamic properties are deprecated: https://php.watch/versions/8.2/dynamic-properties-deprecated. Some tests on PHP 8.2 are failing due to deprecation notices coming from IDS (and the deprecations would occur in general usage of CiviCRM.

The upstream library does not seem to be maintained, and PHP 8.1 patches were applied direct to CiviCRM packages. Therefore I have not opened an upstream PR.

Whilst I was here I also fixed some bad PHPDoc comments.

### DIFF
--- a/IDS/Log/Email.php
+++ b/IDS/Log/Email.php
@@ -74,6 +74,13 @@ class IDS_Log_Email implements IDS_Log_Interface
     protected $headers = null;
 
     /**
+     * Optional envelope string
+     *
+     * @var string
+     */
+    protected $envelope = null;
+
+    /**
      * Safemode switch
      *
      * Using this switch it is possible to enable safemode, which is a spam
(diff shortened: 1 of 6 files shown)


IDS/Log/Email.php
```diff
@@ -160,7 +167,6 @@ protected function __construct($config)
         } elseif (is_array($config)) {
             $this->recipients[]      = $config['recipients'];
             $this->subject           = $config['subject'];
-            $this->additionalHeaders = $config['header'];
```
braders, line 163: `additionalHeaders` was not being used anywhere, so rather than declare the property, I removed this line.

colemanw:
Very unfortunate about the upstream; patching Packages is the only expedient solution so here we go...

PR #364, opened 2023-07-15
[REF][PHP8.2] Resolve PHP8.2 Dynamic property issues by declaring pro…

seamuslee001:
…perties on the class

This aims to resolve the test failure 

```
CRM_Contact_Form_Task_PrintDocumentTest::testPrintDocument
Creation of dynamic property clsTbsZip::$Meth8Ok is deprecated

/home/homer/buildkit/build/build-2/web/sites/all/modules/civicrm/packages/TbsZip/tbszip.php:21
```

### DIFF
--- a/TbsZip/tbszip.php
+++ b/TbsZip/tbszip.php
@@ -17,6 +17,26 @@ define('TBSZIP_STRING',32);    // output to string, or add from string
 
 class clsTbsZip {
 
+  private $Meth8Ok;
+  private $DisplayError;
+  private $Error;
+  private $ArchFile;
+  private $ArchIsNew;
+  private $CdEndPos;
+  private $CdPos;
+  private $CdInfo;
+  private $ArchHnd;
+  private $ArchIsStream;
+  private $AddInfo;
+  private $ReplInfo;
+  private $ReplByPos;
+  private $VisFileLst;
+  private $CdFileNbr;
+  private $CdFileByName;
+  private $CdFileLst;
+  private $LastReadIdx;
+  private $LastReadComp;
+
 	function __construct() {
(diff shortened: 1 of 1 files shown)


demeritcowboy:
I read that as meth now costs $80k. I'll stick with sniffing glue.

Just going to take a quick look what we even use this for and if we still need it - upstream last updated 2014?

colemanw:
Yea but have you seen the price of glue lately?

demeritcowboy:
It seems to be only used in one place in core, and while it could be replaced with regular zip, it has an interesting feature where it can change zips on the fly.

So this seems fine. I was thinking making them private might cause extensions that use it to fail, but the only two that appear to use it have their own versions.

TbsZip/tbszip.php
```diff
@@ -17,6 +17,26 @@
 
 class clsTbsZip {
 
+  private $Meth8Ok;
```
totten, line 20: FWIW, I googled around a little bit. "TBS" stands for "Tiny But Small" (https://www.tinybutstrong.com/) - they seem to have a few inter-related projects (eg https://github.com/Skrol29/tbszip and https://github.com/Skrol29/opentbs and https://github.com/Skrol29/tinybutstrong). The `tbszip.git` hasn't been updated in a long time, but I suspect it's still used - because the other repos have been updated.

Note, eg, OpenTBS was updated 2 months ago - and in that project, the main class [extends clsTbsZip and uses `$this->Meth8Ok`](https://github.com/Skrol29/opentbs/blob/89aea4db66bc76808f70e7fc7bd37980da0f5011/tbs_plugin_opentbs.php#L138C1-L138C89). So my guess is that - to get this upstreamed - it probably needs to be `protected` or `public`.

(cc @Skrol29)

PR #366, opened 2023-08-04
Smarty - Fix e-notice in modifier.escape.php

colemanw:
Overview
-----------
Fixes an ugly e-notice.
FWIW I've also submitted this PR upstream to https://github.com/smarty-php/smarty/pull/892

Before
-------
image

After
------
image


colemanw:
Thanks @demeritcowboy... we'll see what happens with the upstream repo.

PR #367, opened 2023-08-23
kcfinder - Fix cookies when running on alt HTTP port

totten:
Overview
------------

Fix a problem KCFinder where doesn't work with alternate HTTP ports (like `:8000`). Makes it unusable for local testing/triage/maintenance.

Before
---------

* Assigns `cookieDomain` a default value from `HTTP_HOST`
* Console shows multiple issues about setting/reading cookies. Buttons don't work.

Screenshot from 2023-08-22 21-23-37

After
---------

* Assigns `cookieDomain` a default value from `HTTP_HOST`, minus the ports
* OK.

Comments
-------------

I spun up a demo site in case you want to `r-run` the pre-patch code. (<a href="http://site-list.test-1.civicrm.org:8001/?filter=demo-171-*">site info</a>)

### DIFF
--- a/kcfinder/conf/config.php
+++ b/kcfinder/conf/config.php
@@ -101,7 +101,7 @@ $_CONFIG = array(
 
     'mime_magic' => "",
 
-    'cookieDomain' => "",
+    'cookieDomain' => preg_replace(';:\d+$;', '', $_SERVER['HTTP_HOST']),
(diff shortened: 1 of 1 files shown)


seamuslee001:
Tested this on the PR test site which has a custom port and confirmed KCFinder works there

PR #337, opened 2021-12-14
Update rcube_html2text.php to the latest version from RoundCube 1.5.1

homotechsual:

PR #369, opened 2023-08-30
Update rcube_html2text.php to upstream version

larssandergreen:
Once every ten years or so.
Still maintained. Motivation was to remove the `<b>` to UPPERCASE conversion, which they took out a few years back.

### DIFF
--- a/html2text/rcube_html2text.php
+++ b/html2text/rcube_html2text.php
@@ -1,16 +1,10 @@
 <?php
 
 /**
- * We have switched to using the roundcube version since it appears to
- * be maintained and updated compared to the original version
- *
- * CRM-12766
- */
-
-/**
  +-----------------------------------------------------------------------+
  | This file is part of the Roundcube Webmail client                     |
- | Copyright (C) 2008-2012, The Roundcube Dev Team                       |
+ |                                                                       |
+ | Copyright (C) The Roundcube Dev Team                                  |
  | Copyright (c) 2005-2007, Jon Abernathy <jon@chuggnutt.com>            |
  |                                                                       |
  | Licensed under the GNU General Public License version 3 or            |
@@ -104,6 +98,11 @@
  */
 class rcube_html2text
 {
+    const LINKS_NONE = 0;
+    const LINKS_END = 1;
+    const LINKS_INLINE = 2;
+    const LINKS_DEFAULT = self::LINKS_END;
+
     /**
      * Contains the HTML content to convert.
      *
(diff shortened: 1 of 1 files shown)


demeritcowboy:
jenkins add to whitelist

demeritcowboy:
There was a PR last year that just never got finished. The same test might fail: https://github.com/civicrm/civicrm-packages/pull/337

eileenmcnaughton:
The test fail is pretty closely linked to the discussion & the agreed resolution still stands (it was just never done)

eileenmcnaughton:
If this fails on related tests we should merge it & then fix the tests - cos they are in a different repo & we have to do one before the other!

eileenmcnaughton:
Fails are an older phpunit version being called...

Civi\FlexMailer\FlexMailerSystemTest::testBasicHeaders
Error: Call to undefined method Civi\Test\EventCheck@anonymous::assertMatchesRegularExpression()

seamuslee001:
Jenkins re test this please

larssandergreen:
OK, so it turns out that this file is not actually in use at all: https://github.com/civicrm/civicrm-core/pull/23971

Can we just get rid of this entirely? Per @demeritcowboy, no [other use in universe.](https://github.com/civicrm/civicrm-core/pull/23971#issuecomment-1179421632)

demeritcowboy:
Ha, I knew there was something about this bugging me. And it would explain why the test from earlier didn't fail this time.

larssandergreen:
Yeah, the lack of failure clued me in.

larssandergreen:
BTW, test failures were mostly the regex function upgrades for PHPUnit9, it seems that some of the tests here are still running older versions of core. Maybe that's how it is supposed to work? I think it uses [PHPUnit8 for pre-5.65.](https://github.com/civicrm/civicrm-buildkit/blob/e9b60cdf87d192496673f296fb55a0a447f0c33b/bin/civi-test-run#L23)

Also a couple failures about how it couldn't send a mailing in staging:
```
Civi\FlexMailer\ConcurrentDeliveryTest::testBatchActivityTargets with data set #1 (false)
Failure in api call for job process_mailing:  Job has not been executed as it is a Staging (non-production) environment.
Failed asserting that a integer is empty.
```

larssandergreen:
Here are the todos [from a year ago](https://lab.civicrm.org/dev/core/-/issues/3676#note_78336):

Remove from packages
Remove from Utils/Check/Component/Security.php

larssandergreen:
Here's a PR to remove from packages, though I expect the same failures:
https://github.com/civicrm/civicrm-packages/pull/370

PR #372, opened 2023-09-23
Move our security for fetch in smarty 2 to hack smarty rather than an override

eileenmcnaughton:
… 

Normally I would favour a CRM fix over hacking a package but in this case the CRM fix is a blocker to us becoming less hacked (ie getting off smarty 2) because the smarty 3 signature differs from smarty2


colemanw:
I see you went down the linting rabbit hole here...

PR #374, opened 2023-12-11
dev/joomla#46 - Fixed KC Finder Not Working Under Joomla 4

crusonweb:
Updated the authentication process for Joomla 4 in KCFinder as described in GitLab (https://lab.civicrm.org/dev/joomla/-/issues/46).  The problem that needed to be fixed is that the Browse Server button will throw a 500 error under Joomla 4.  This is because Joomla changed JFactory::getApplication, so it doesn't return the same values as it did under previous versions.

### DIFF
--- a/kcfinder/integration/civicrm.php
+++ b/kcfinder/integration/civicrm.php
@@ -200,10 +200,25 @@ function authenticate_joomla($config) {
   require_once ( JPATH_BASE .DS.'includes'.DS.'defines.php' );
   require_once ( JPATH_BASE .DS.'includes'.DS.'framework.php' );
 
-  $mainframe = JFactory::getApplication('administrator');
-  $mainframe->initialise();
+  if (version_compare(JVERSION, '4.0.0', 'lt')) {
+    $mainframe = JFactory::getApplication('administrator');
+    $mainframe->initialise();
+    
+    $user_id = JFactory::getUser()->id;
+  } else {
+    // Boot the DI container.
+    $container = \Joomla\CMS\Factory::getContainer();
 
-  if (JFactory::getUser()->id == 0) {
+    // Alias the session service key to the web session service.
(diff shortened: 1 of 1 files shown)


vingle:
Have tested on Joomla 4 and can confirm the KC Finder on 4 is broken before this, and loads afterwards.

Only question - I've not used KCFinder in JoomlaCivi - but should there be an 'insert image' button/icon on this screen? Maybe it's permissions but I couldn't see how to get from here back to the insert image dialog.

image

crusonweb:
I don't recall there being an insert image button in the past and I checked a couple of different sites (including an old Drupal one) and I don't see a button on them.  I have always just double clicked the image and it takes you back to the insert image dialog with the field populated with the image path.

vingle:
@crusonweb I tried double click but it didn't work, maybe a temp browser issue / something else. Overall tho this PR turns something from giving a 500 error, with its intimidating Joomla red screen below, to the screengrab above. So this PR has my vote.

image

crusonweb:
Thank you!

aydun:
Error is 
```
11 12 2023 13:57:17.268:ERROR [launcher]: Cannot start ChromeHeadless
	Can not find the binary /nix/var/nix/profiles/bknix-dfl/bin/chromium
	Please set env variable CHROME_BIN
```

Let's see if it can find another time around...

Jenkins test this please.

aydun:
Test this please

vingle:
@aydun - is Jenkins able to test a Joomla build? It's the Drupal Test that's failing…?

There's a new Joomla bug related to this - https://lab.civicrm.org/dev/joomla/-/issues/57 - so it would be great to figure out a way to get this merged. I'm not sure what more I can do to help, as I've already tested and said it's good / big improvement.

aydun:
Let's try again so we can see what the fail is

Test this please

aydun:
Test error appears unrelated.  Merging on @vingle's testing.

PR #375, opened 2023-12-17
Migrate Snappy to composer

bb:
See https://github.com/civicrm/civicrm-core/pull/28673 for corresponding change in core.


demeritcowboy:
jenkins ok to test

PR #378, opened 2024-01-05
Fix problem with file names.

spalmstr:
When running the latest version of CiviCRM (5.70alpha1) under IIS/Windows that some reports would not run because the system could not write a compiled template file. Investing this revealed that the process involves writing to a temporary file and then renaming it. This last step was failing because the destination file name contained line feeds or carriage returns. These were being put in by the getBasename function one or both of the changed files. This PR adds a preg_replace step that removes the offending characters (and tabs). Reports will now run.

It may be that there is a better fix, but this is at least a stopgap.

### DIFF
--- a/smarty3/vendor/smarty/smarty/libs/sysplugins/smarty_internal_resource_file.php
+++ b/smarty3/vendor/smarty/smarty/libs/sysplugins/smarty_internal_resource_file.php
@@ -89,6 +89,7 @@ class Smarty_Internal_Resource_File extends Smarty_Resource
         // Some basenames contained carriage returs or new lines. At least
         // under Windows, this leads to illegal file names. The regular
         // expression removes new lines, carriage returns and tabs.
+    
         $basename = basename($source->filepath);
         return preg_replace( '/(\r\n)+|\r+|\n+|\t+/', '', $basename );
     }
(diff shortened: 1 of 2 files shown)


spalmstr:
I don't understand the failure that says No Changes as two files are changed.

demeritcowboy:
@spalmstr I wonder if it's better to add a line `$tplSource = trim($tplSource);` just before this line: https://github.com/civicrm/civicrm-core/blob/cb64f2a1187fed32161ac2319bf45e3acbcca99f/CRM/Core/Form/Renderer.php#L238 in core? Just to minimize patches to 3rd party libs.

demeritcowboy:
Alternate PR: https://github.com/civicrm/civicrm-core/pull/28904

demeritcowboy:
@spalmstr Note an alternate alternate https://github.com/civicrm/civicrm-core/pull/28909 has been merged - if that also fixes the issue you're seeing we could close this?

spalmstr:
@demeritcowboy https://github.com/civicrm/civicrm-core/pull/28909 appears to fix the issue, too. When I reverted this PR in my local test system and applied the alternate, the error did not recur. So, this PR might as well be closed.

PR #379, opened 2024-01-05
Allow more dynamic properties

spalmstr:
Later versions of PHP deprecate having dynamic variables in the class constructors and CiviCRM complains when they are used. This PR allows the 'offending' code to use dynamic variables in the constructors, since, without knowing the code, it is possible that the same variable will have different properties depending on how it is called. Obviously, it would be better to declare them if one knew how it was being used.


eileenmcnaughton:
@spalmstr I've been digging a bit on Smarty & looks like we can push through to Smarty 4 - which hopefully doesn't have any remaining php compatibility issues - https://lab.civicrm.org/dev/core/-/issues/4954

eileenmcnaughton:
We killed Smarty3 ! Smarty 5 is OK with later php

PR #380, opened 2024-01-27
Add Smarty4 folder

eileenmcnaughton:
I tried out smarty4& it looks like it's a small step from Smarty3 to Smarty4 but a bit trickier to Smarty5 until we get rid of Smarty2

I think it makes sense to add Smarty4 to our vendor dir to try out in the same way as Smarty3 & when we hard switch my recent testing suggests 4 rather than 3 makes sense


eileenmcnaughton:
test this please

eileenmcnaughton:
test this please

colemanw:
Why are we doing this in `civicrm-packages` instead of `civicrm-core/composer.json`?

demeritcowboy:
Probably because then the autoloader will pick it up even if you haven't enabled smarty 4.

colemanw:
Ok

PR #381, opened 2024-02-01
dev/core#4955 Hack Smarty3+ getTemplateVars into Smarty2

eileenmcnaughton:
Per https://lab.civicrm.org/dev/core/-/issues/4955 overriding this function in a core compatibility class allows us to upgrade to Smarty3 & 4 but we hit a brick wall at 5.

Moving it into the package we plan on removing soon gets us past this (I will remove from the core SmartyCompatibily once this is merged).

Not sure if there are other functions - this is the blocking one ATM

### DIFF
--- a/Smarty/Smarty.class.php
+++ b/Smarty/Smarty.class.php
@@ -1068,6 +1068,22 @@ class Smarty
         }
     }
 
+  /**
+   * Returns a single or all template variables
+   *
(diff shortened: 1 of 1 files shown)


colemanw:
Not sure what to make of those tests but they seem to be failing for every PR in this repo.

PR #384, opened 2024-02-06
Add clearAssign transitional shim

eileenmcnaughton:


demeritcowboy:
It looks like clearAssign used to accept 0 params. I'm seeing crashes now from https://github.com/civicrm/civicrm-core/blob/cba6d4b825b7343c9274cd46b1013999ff3f010a/CRM/Core/Smarty.php#L389

demeritcowboy:
Oh wait the problem is actually that line - it was added recently.

PR #388, opened 2024-02-15
Add load filter compatibility function to Smarty2

eileenmcnaughton:
We originally put these in SmartyCompatibility but by now all except this have been transferred to be hacks on Smarty in packages so lets move this too


seamuslee001:
Test fail unrelated

PR #395, opened 2024-04-05
Port one more function to Smarty2

eileenmcnaughton:


totten:
Follow-up to https://github.com/civicrm/civicrm-core/pull/29851 which seems to fix the sporadic failure when running v2. Merge on pass.

PR #397, opened 2024-05-07
Fix setTemplateDir to actually work

colemanw:
Overview
-----

Fix broken backported function to support https://github.com/civicrm/civicrm-core/pull/30148

Before
------
Function creates garbage.

After
------
Function does what it's supposed to do.

### DIFF
--- a/Smarty/Smarty.class.php
+++ b/Smarty/Smarty.class.php
@@ -1107,8 +1107,8 @@ class Smarty
    *
    * @return \Smarty current Smarty instance for chaining
    */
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
yeah - I guess this one doesn't need to work with the 'right' signature since it's hidden away in Smarty2

demeritcowboy:
Does this need a 5.73 port?

colemanw:
@demeritcowboy this ended up not being on the critical path for anything.

PR #398, opened 2024-05-08
Backport addPluginsDir to Smarty2

eileenmcnaughton:


totten:
There are some downsides to backporting in this context -- it muddies the waters for contrib. From POV of an extension-developer who's figuring out how to work with different versions...

* You're not just thinking about "Smarty 4" vs "Smarty 2"...
* You have to think about "Smarty 4" vs "CiviCRM 5.63's flavor of Smarty 2" vs "CiviCRM 5.69's flavor of Smarty 2" vs "CiviCRM 5.75's flavor of Smarty 2."
* If you want to test your extension with Smarty 2+5, you can't just switch Smarty 2<=>5. You actually have to test the different releases of Civi.
* Ultimately, for an extension-developer who needs to support a few versions of `civicrm-core`, they shouldn't rely on this backport.

On the plus side... 

* The patch doesn't look breaky on its own terms.
* The patch saves ~4 lines of code that would otherwise be needed in `CRM_Core_Smarty`. (`if (is_callable([$smarty, 'addPluginsDir']))...`)

eileenmcnaughton:
Yeah - I took a look at @totten's comments -but we have already made the call to backport into the outgoing package with a bunch of columns so I think that's fine - we will be over this transition soon-ish

PR #399, opened 2024-05-08
Add getVersion function

eileenmcnaughton:
in conjunction with https://github.com/civicrm/civicrm-core/pull/30156 this stops core from attempting to add a plugin - which is otherwise dealt with


totten:
civibot, test this please

eileenmcnaughton:
I think this got missed in the related merges - not strictly part of the fix in the end but kinda part of the various fixes we try

PR #400, opened 2024-05-08
Smarty5 start using an extension for Smarty5

eileenmcnaughton:
This is how it would work - although I started with the easiest - ie modifier  - the others do not seem to offer a compiler


seamuslee001:
@eileenmcnaughton where are we at with this one?

PR #408, opened 2024-06-12
kcfinder - Add support for Standalone

totten:
ping @ufundo 

Steps to reproduce
--------------------

1. Install CiviCRM standalone (*esp. standard/tar-style layout*)
2. Create/update `core/civicrm.config.php`.  Depending on which PRs/layouts you're working with you, the content of this file may change. For example, if using current revision of @ufundo's 29960 with the simpler (non-composer) layout, then the file would need to say:
    ```php
    <?php
    chdir(dirname(__DIR__));
    require_once dirname(__DIR__) . '/civicrm.standalone.php';
    ```
3. Login
4. Go to "New Mailing" or "New Activity". Try to "Browse server" and/or "Upload" ("Send to server").

Before
-------

Screenshot 2024-06-12 at 8 29 58 AM

After
-----

Screenshot 2024-06-12 at 8 31 53 AM

Comments
------------

* Noticed this problem while trying to evaluate/confirm file-layouts.
* This is the same permisison-check used on WP, D7, D8+, BD.
* Yes, yes, of course `kcfinder` is the Spawn of Satan, the Destroyer of Worlds, the Eye-Burning Dragon, the Great Purveyor of Tortuous 1990's Iconography. If you can replace it with a 5 minute patch, great. But realistically, it's what we have right now. It's better if it doesn't crash. 😃

### DIFF
--- a/kcfinder/core/class/browser.php
+++ b/kcfinder/core/class/browser.php
@@ -157,7 +157,7 @@ class browser extends uploader {
             $this->sendDefaultThumb();
 
         $dir = $this->getDir();
-        $file = "{$this->thumbsTypeDir}/{$_GET['dir']}/${_GET['file']}";
+        $file = "{$this->thumbsTypeDir}/{$_GET['dir']}/{$_GET['file']}";
 
         // Create thumbnail
         if (!is_file($file) || !is_readable($file)) {
--- a/kcfinder/integration/civicrm.php
+++ b/kcfinder/integration/civicrm.php
@@ -69,6 +69,9 @@ function checkAuthentication() {
     case 'Drupal8':
       $auth_function = 'authenticate_drupal8';
       break;
+    case 'Standalone':
+      $auth_function = 'authenticate_standalone';
+      break;
     }
     if(!$auth_function($config)) {
       throw new CRM_Core_Exception(ts("You must be logged in with proper permissions to edit, add, or delete uploaded images."));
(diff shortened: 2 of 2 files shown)


totten:
@artfulrobot Yeah, I'm not sure the answer to that.

* Traditionally, in other UF's, the `civicrm.config.php` was generated by `GenCode` and then packaged into the zip/tarball. Its behavior was to boot Civi and then UF. Its main appeal was that (*if your code lived under the core source tree*) you could find it in a predictable place (eg `../civicrm.config.php`).
* Given all the other scaffolding bits, it probably could be added in a similar way.
* Obviously, it's not a great file. 
* One could potentially patch `kcfinder/integration/civicrm.php` to [check another location/relpath](https://github.com/civicrm/civicrm-packages/blob/5.75/kcfinder/integration/civicrm.php#L42-L46). Maybe go up one more dir and look for `civicrm.standalone.php`. 
* I'm not sure what's happening for KCFinder in D8/9/10. That might have some other way which doesn't require `civicrm.config.php`.

ufundo:
> Maybe go up one more dir and look for civicrm.standalone.php

👍👍

The idea of that file was to provide a common boot path for cv , index.php, and anything else. So that would be my preferred solution. 

I can give that a go this week.

ufundo:
> I'm not sure what's happening for KCFinder in D8/9/10. That might have some other way which doesn't require civicrm.config.php.

Was also stumped by this. That filename looks pretty hardcoded on the kcfinder side, but doesnt seem to exist on a D10 install. Some Drupal magic??

ufundo:
Thanks @totten .

Have raised https://github.com/civicrm/civicrm-packages/pull/410 as the easy win.

Though did go a bit round the houses thinking about civicrm.config.php ... will add a comment on that PR...

PR #410, opened 2024-07-01
KCFinder - look for civicrm.standalone.php as well as civicrm.config.php

ufundo:
Currently KCFinder is broken on Standalone because Standalone doesn't have a `civicrm.config.php` file, which the integration uses to boot CiviCRM to authenticate the user.

`civicrm.standalone.php` does the equivalent job - this patch looks for that as an alternative.

It searches up parent directories, which seemed the neatest way to support:
a) the existing case of `civicrm.config.php` in the civicrm-core folder, 
b) `civicrm.standalone.php` being in any parent directory (civicrm-core could be `core` or `vendor/civicrm/civicrm-core` relative to `civicrm.standalone.php`)
c) conceivable future permutations involving civicrm.config.php in a parent directory?

I wasn't sure why/if any of the previous `chdir` was still necessary - removing works for me but @totten do you know?

See also discussion on https://github.com/civicrm/civicrm-packages/pull/408

Note: KCFinder currently similarly broken out-of-box for D8+, (C) could be useful for fixing that

### DIFF
--- a/kcfinder/integration/civicrm.php
+++ b/kcfinder/integration/civicrm.php
@@ -36,13 +36,32 @@
  *
  */
 
+
 function checkAuthentication() {
   static $authenticated;
   if ( !isset( $authenticated ) ) {
-    $current_cwd   = getcwd();
-    $civicrm_root  = dirname(dirname(getcwd()));
     $authenticated = false;
-    require_once "{$civicrm_root}/civicrm.config.php";
+
+    // used to chdir at the end of this function - not sure if necessary?
+    $current_cwd = getcwd();
+
+    function findConfigFile(string $search_path): string {
+      while ($search_path) {
+        foreach(['civicrm.config.php', 'civicrm.standalone.php'] as $config_filename) {
+          $config_file_candidate = $search_path . DIRECTORY_SEPARATOR . $config_filename;
+
+          if (file_exists($config_file_candidate)) {
+            return $config_file_candidate;
+          }
+        }
+
+        $search_path = dirname($search_path);
+      }
+
+      throw new \Exception('KCFinder couldn\'t find civicrm.config.php or civicrm.standalone.php to check authentication');
+    }
+
+    require_once findConfigFile(__DIR__);
     require_once 'CRM/Core/Config.php';
 
     $config = CRM_Core_Config::singleton();


ufundo:
This seems the easiest fix, but I do wonder about `civicrm.config.php` more generally...

It is used in a few other external endpoints ( like cron.php ), which are think are generally seen as a bit legacy. But if we did want to support I think a relatively straightforward alternative would be https://github.com/civicrm/civicrm-core/commit/ff5f5b89462dd902042e4d6d50769db3b1628051

I do find it a bit awkward having the `civicrm.config.php` added in to the core folder on a per-UF basis though - and it's difficult for composer-installs (composerfied Standalone or D8+).

So my preferred solution might be:
- replace require_once ../../civicrm.config.php with search up the directory tree to find civicrm.config.php (i.e. this PR, but without the reference to `civicrm.standalone.php` )
- generate the civicrm.config.php alongside civicrm.standalone.php in the root of the Standalone install

Extra effort, but could be worth it if also useful for the newer alternative entrypoints like `iframe.php`?

@totten

kcfinder/integration/civicrm.php
```diff
@@ -36,13 +36,32 @@
...
 function checkAuthentication() {
   static $authenticated;
   if ( !isset( $authenticated ) ) {
-    $current_cwd   = getcwd();
-    $civicrm_root  = dirname(dirname(getcwd()));
     $authenticated = false;
-    require_once "{$civicrm_root}/civicrm.config.php";
+
+    // used to chdir at the end of this function - not sure if necessary?
+    $current_cwd = getcwd();
```
totten, line 46: Oh, that is funny.

+1 for what you've done here - keep for now, but flag it to see as a curiosity.

* FWIW, it is plausible that this is needed. The effect of loading `civicrm.config.php` is to bootstrap some mix of Civi+UF (BD/D7/D10/J/WP) -- and it's quite possible that *some* combination of Civi+UF has the side-effect of changing CWD (*to match the UF's project-root or HTTPD web-root*).

* Our E2E test-coverage over this bit is non-existent. (The more we change this file, the more we have to think about QA work.)

totten:
> (@ufundo) This seems the easiest fix, but I do wonder about civicrm.config.php more generally...

I like you're thinking process...

> It is used in a few other external endpoints ( like cron.php ), which are think are generally seen as a bit legacy.

Grepping on `civicrm-core` and `civicrm-packages`, I think most have better options in the regular router now. The ones which are interesting:

* `kcfinder.php`: OK, this one. :)
* `bin/cron.php`: Already has alternatives within standalone (CLI: `cv api`, HTTP: `civicrm/ajax/rest`), so it isn't strictly required. The only thing... for folks who need HTTP-based triggering, the [Sysadmin Docs](https://docs.civicrm.org/sysadmin/en/latest/setup/jobs/) push them to `cron.php`.  From their POV, using HTTP-REST for cron is a long adventure.
    * (*Wishlist: the UX for configuring either HTTP-cron.php or HTTP-REST feels unnecessarily brutal... and it could use a do-over...*)
* `bin/regen.sh` (specifically `sql/GenerateData.php`): This is still using `civicrm.config.php`, and it would be good to support on standalone.
    * There are some long-standing desires to reorganize this sample-data workflow, but... we haven't yet. And that's not fast.
    * Fast alternative: change [this](https://github.com/civicrm/civicrm-core/blob/181ffae248d71881caf21cab9cda7a88f59b9954/bin/regen.sh#L46) to `cv scr sql/GenerateData.php` and then drop the call to `require_once civicrm.config.php`. (I think it's fairly safe to expect anyone using `regen.sh` to have `cv`.)

Aside from kcfinder, I don't think any of the others really *need* to go through `civicrm.config.php`(/`civicrm.standalone.php`). But I do agree that reviving `civicrm.config.php` is a perfectly legit way to address all three.

> I do find it a bit awkward having the civicrm.config.php added in to the core folder on a per-UF basis though - and it's difficult for composer-installs (composerfied Standalone or D8+).

Yup. Both good reasons for its infamy.

Another: on WP, some site-builders can (*reportedly*) scramble the file-layout in ways that make it very hard to use a folder-search.

>  could be worth it if (*`civicrm.config.php` were*) also useful for the newer alternative entrypoints like `iframe.php`?

Yes...but no. :) From my POV, the design-direction of `iframe.php` is an upgrade over the `civicrm.config.php` design.

For `iframe.php`, the public interface (URL) is completely decoupled from the internal/physical layout of Civi+UF, so it's easier to preserve inbound link compatibility over time. It boots CMS-first, so it is consistent with regular pageviews -- but also allows detailed influence over CMS-boot options. And it should adapt more readily to some wickedly scrambled file-layouts -- as can happen with customized D8+ and customized WP. (It may compromise a bit on compatibility for some cookie-cutter/locked-down webhosts.)

-------------

OK, I've tested this PR by uploading images in kcfinder (with a different techniques). It works on the local standalone site and Jenkins' D7 site.

Flagging merge-ready. If you still like this approach, go for it.

I also think your preferred approach sounds good. If you'd rather go down that path, I'll test it too.

PR #416, opened 2024-11-06
WIP: POC fix for dev/core#5571.

twomice:
Reference issue: https://lab.civicrm.org/dev/core/-/issues/5571

- This solves the stated problem (upon validation of required checkbox/radio fields), and my own initial, manual, testing shows it doesn't break validation of required fields of other types.

Other thoughts:

- I'm unfamiliar with the `packages/` repo. Is a PR here even a reasonable thing to do?
- Adding params to a method doesn't "feel great", but this is just a POC. 
- I'm unsure whether an equivalent fix would be relevant for similar method in these places:
  - `packages/_ORIGINAL_/HTML/QuickForm/Renderer/ITStatic.php`
  - `packages/_ORIGINAL_/HTML/QuickForm/Renderer/ArraySmarty.php`
  - `packages/HTML/QuickForm/Renderer/ITStatic.php`

### DIFF
--- a/HTML/QuickForm/Renderer/ArraySmarty.php
+++ b/HTML/QuickForm/Renderer/ArraySmarty.php
@@ -197,7 +197,7 @@ class HTML_QuickForm_Renderer_ArraySmarty extends HTML_QuickForm_Renderer_Array
             $this->_renderRequired($ret['label'], $ret['html'], $required, $error);
         }
         if ($error && !empty($this->_error)) {
-            $this->_renderError($ret['label'], $ret['html'], $error);
+            $this->_renderError($ret['label'], $ret['html'], $error, $element->_type);
             $ret['error'] = $error;
         }
         // create keys for elements grouped by native group or name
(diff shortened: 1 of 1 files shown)


totten:
> I'm unfamiliar with the packages/ repo. Is a PR here even a reasonable thing to do?

It depends on the package. For HTML_QuickForm, yes. Upstream hasn't released in years.

> I'm unsure whether an equivalent fix would be relevant for similar method in these places:
>
> * packages/_ORIGINAL_/HTML/QuickForm/Renderer/ITStatic.php
> * packages/_ORIGINAL_/HTML/QuickForm/Renderer/ArraySmarty.php

These files (`_ORIGINAL_`) are just copies of the original/upstream repo. They don't need to be edited.

> * packages/HTML/QuickForm/Renderer/ITStatic.php

As far as I can tell, there's nothing in core or universe that calls out to `ITStati`c. TBH, we could just as well delete half the items in `HTML/QuickForm/Renderer/`...

> Adding params to a method doesn't "feel great", but this is just a POC.

I can see how it might feel awkward -- the name suggests a kind of central function. But I think I agree it's OK. When I dig around for references to this method, it really does seem like (1) it's not a substantive part of the inheritance hierarchy and (2) it's only got the one internal caller.

------------

However, I'm finding the change to be regressive elsewhere, i.e.

* Setup your test as in the description
* Note the profile
* Open the profile as a standalone page - eg `civicrm/profile/edit?gid=12&reset=1`
* Fill in everything - but leave the checkboxes all blank
* Submit the profile.

For the standalone profile page, it worked before the patch. But with the patch, checkboxes go MIA. Compare:

* Before/Expected
    Screenshot from 2025-01-02 16-29-16

* After/MIA
    Screenshot from 2025-01-02 16-27-56

eileenmcnaughton:
I've added a variant of this here - https://github.com/civicrm/civicrm-core/pull/31705/files#diff-5a429c6ee446ea34fcc58e2cf1e84a8befea3239e8519ebe59b32d335761b836R87 

it is probably a bit narrow in that it only works on ones with the options_in_line configured - but it not in packages - still thinking there is another soluttion tho

seamuslee001:
@twomice @eileenmcnaughton is this still needed?

PR #417, opened 2025-02-04
Fix php8.x support

eileenmcnaughton:
This is generally not an issue cos this code is possibly not hit - or not in tested/ frequent code

However, it does get parsed by linters which was where we hit it. Note the package is thoroughly abandonded but is referenced from the DBObject PEAR class so we are still stuck with it

https://pear.php.net/package/Date/:


demeritcowboy:
I think can merge this without the tests. Tests are obviously not running anywhere right now.

PR #422, opened 2025-04-01
Fix PHPgettext fseek deprecation warnings in PHP 8.3

gnodet:
In PHP 8.3, passing null as the offset parameter to fseek() is deprecated. This commit adds the SEEK_SET parameter to the fseek() calls in the PHPgettext streams.php file to avoid these deprecation warnings.

The warning that appears without this fix is:
"fseek(): Passing null to parameter #2 () of type int is deprecated"

### DIFF
--- a/PHPgettext/streams.php
+++ b/PHPgettext/streams.php
@@ -105,7 +105,7 @@ class FileReader {
 
   function read($bytes) {
     if ($bytes) {
-      fseek($this->_fd, $this->_pos);
+      fseek($this->_fd, $this->_pos, SEEK_SET);
 
       // PHP 5.1.1 does not read more than 8192 bytes in one fread()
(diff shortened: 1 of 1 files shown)


colemanw:
@gnodet thanks for working on this.

I think there's a problem though: according to the [php documentation](https://www.php.net/manual/en/function.fseek.php), `fseek` already defaults to `SEEK_SET` as the 3rd param, so this PR doesn't actually change anything. Omitting it is the same as adding it.

gnodet:
Fair enough, not sure how I got the warning on the web page though...

gnodet:
image

I still have the following when running the plain docker compose example.

demeritcowboy:
@gnodet Can you debug what $filename is on line 91. It seems like it's trying to read a file that doesn't exist.

PR #441, opened 2026-04-22
remove git-footnote

MegaphoneJon:
Given a Jira issue number, this adds a Jira URL to the commit message.

Something tells me we don't need this anymore.


colemanw:
I can live without it. Thanks @MegaphoneJon

PR #446, opened 2026-07-17
Prevent retry of single statement when within db transaction

larssandergreen:
The retry logic here (which we added ourselves many years ago) works fine for a single statement, but for a transaction with multiple statements, it can result in data being lost. What happens is if we have a few statements and one of the later ones hits a deadlock, it rollsback and then retries just that statement. If that retry succeeds, then that single statement gets committed without the earlier statements. The result is inconsistent data.

In our case, we were updating a ContributionRecur and then sending an email via queue inside a transaction, and the second hit a deadlock and then succeeded on retry, so the email confirming recurring cancellation was being sent but the update to cancel the recurring donation wasn't happening. But I don't see why this couldn't also happen for transactions in core like Contribution plus LineItem etc or in a Contact merge.

This changes the logic so we only retry if we aren't inside a transaction (the check is the same one CRM_Core_Transaction uses). For completeness, we also check if innodb_rollback_on_timeout=1 (rollback the transaction on 1205), which would cause the same behaviour. We no longer need $firstError because we're either throwing immediately or retrying a single statement (I suppose the error for the single statement could change on a subsequent retry, but I think throwing the most recent one is fine in that case as it would be a legit reason).

I tested this out with some toy test scenarios to confirm the behaviour change.

### DIFF
--- a/DB/DataObject.php
+++ b/DB/DataObject.php
@@ -2696,22 +2696,26 @@ class DB_DataObject extends DB_DataObject_Overload
             $result = $DB->query($string);
           }
           catch (PEAR_Exception $e) {
-            if ($tries == 0) {
-              // The original sin was what triggered the retry. Sometimes the retry fails because mysql has done an internal rollback
-              // of previous queries in the transaction so it has essentially failed to recover from the deadlock. If we can't
-              // recover we should return the original error.
-              $firstError = $e;
-            }
-            // CRM-21489 If we have caught a DB lock - let it go around the loop until our tries limit is hit.
-            // else rethrow the exception. The 2 locks we are looking at are mysql code 1205 (lock) and
-            // 1213 (deadlock).
+            // CRM-21489 If we have caught a DB lock - let it go around the loop until our tries limit is hit,
+            // if it is a single statement outside a transaction, else rethrow the exception.
+            // The 2 locks we are looking at are mysql code 1205 (lock) and 1213 (deadlock).
             $dbErrorMessage = $e->getCause()->getUserInfo();
-            if (!stristr($dbErrorMessage, 'nativecode=1205') && !stristr($dbErrorMessage, 'nativecode=1213')) {
-              throw $firstError;
+            $isDeadlock = stristr($dbErrorMessage, 'nativecode=1213');
+            $isLockTimeout = stristr($dbErrorMessage, 'nativecode=1205');
+            if (!$isLockTimeout && !$isDeadlock) {
+              throw $e;
             }
-            $message = (stristr($dbErrorMessage, 'nativecode=1213') ? 'Database deadlock encountered' : 'Database lock encountered');
+            if (\Civi\Core\Transaction\Manager::singleton()->getFrame() !== NULL) {
+              // A 1213 deadlock rolls back the whole transaction, not just the failed statement
+              // (as does a 1205 lock timeout when innodb_rollback_on_timeout=1)
+              // Retrying just the failed statement will commit it without the transaction's earlier statements, so throw instead.
+              if ($isDeadlock || $DB->getOne('SELECT @@innodb_rollback_on_timeout')) {
+                throw $e;
+              }
+            }
+            $message = ($isDeadlock ? 'Database deadlock encountered' : 'Database lock encountered');
             if (($tries + 1) === $maxTries) {
-              throw new CRM_Core_Exception($message, 0, array('sql' => $string, 'trace' => $firstError->getTrace()));
+              throw new CRM_Core_Exception($message, 0, array('sql' => $string, 'trace' => $e->getTrace()));
             }
             CRM_Core_Error::debug_log_message("Retrying after $message hit on attempt " . ($tries + 1) . ' at query : ' . $string);
             continue;


totten:
> This changes the logic so we only retry if we aren't inside a transaction 

OK, less crazy. 😃 🎉  

Aside, non-blocker, part that I don't understand -- when is this "retry" handler actually useful?

* If you're inside a transaction, then it's not safe to use this kind of retry (as described above). OK, so now it's disabled.
* If you're outside a transaction, then... there's no row-locking, right? But if there's no row-locking, then ___how do you get deadlocks?___ When is the naive-retry useful?

I guess it's possible to get deadlocks on other things (e.g. table-locking or user-managed locking), but I don't think there's as much of this going on in Civi. But maybe.

It'll be interesting to watch your logs in prod. Do you actually get more/lesser/same %interactions with retries? Do you get more/lesser/same %interactions with deadlock-failures?

seamuslee001:
cc @johntwyman

demeritcowboy:
> how do you get deadlocks

Besides just heavy resource usage, where a retry won't help, I've seen deadlocks with indexes. Like in a db with lots of dedupe rules configured, and say 1 million contacts, an insert/update of a contact causes it to update 20 indexes on civicrm_contact, which then blocks. A retry might help there, but in that specific case the real thing to do is get rid of any unneeded indexes. Anyway, there's one example.

larssandergreen:
@totten I think you can still get row (or gap) locks from single statements if they write.
I think in our particular case, it's the heavy traffic inserting a bunch of queue items while runners delete them that's the issue. It wasn't even that rare, about 0.4% of these recent attempts appear to have failed in this way.

We can definitely do some log analysis later to look at trends.

larssandergreen:
It looks like using SKIP LOCKED would help for the queue table specifically, but that would require MySQL 8.

totten:
> > > (totten) ... how do you get deadlocks...
> >
> > (demerit) ... I've seen deadlocks with indexes...
>
> (lars) I think you can still get row (or gap) locks from single statements if they write.

Aaah, interesting. So MySQL will make *an implicit transaction* around a single statement... to coordinate index updates. Presumably, the implicit transaction covers other things -- like triggers and `log_*` table updates. (*Which wouldn't matter for queues; but could matter for Contacts and other business-entities.*) In any case, that makes sense. I can imagine how there may be some re-triable single-statement deadlocks.

> (demerit) the real thing to do is get rid of any unneeded indexes.

That path scare me -- because the "need" is subjective. It implies need-based/dynamic/configurable indexes -- which means more possible configurations to wrap our heads around. (And I don't think we have a practice of making indexes speculatively -- most indexes exist because *someone* thought it was needed.)

I feel like we're missing something -- something higher level -- something in our conception of "unit of work" -- that would make it easier to schedule/retry these chunks (regardless of whether indexes exist)...

> (lars) It looks like using SKIP LOCKED would help for the queue table specifically, but that would require MySQL 8.

OK. Well, MySQL 8 could use different driver-class (eg `CRM_Queue_Queue_Sql57` vs `CRM_Queue_Queue_Sql80`). Both MySQL versions are used by nightly test-runs. We just need to be sure that the classes are enforcing comparable behavior.

(*The current [QueueTest](https://github.com/civicrm/civicrm-core/blob/9f3c06dfeadfad5e28bde510cd2797f2b4c8773c/tests/phpunit/CRM/Queue/QueueTest.php#L40-L57) does some QA, but I'm not sure how sharply it tracks the nuances of `Sql{57,80}` vs `SqlParallel{57,80}`. And IIRC `QueueTest` doesn't spawn concurrent agents. The [`coworker` tests](https://lab.civicrm.org/dev/coworker/-/tree/master/tests?ref_type=heads) do spawn more agents, but I don't think they look to the nuances of `Sql` vs `SqlParallel`.*)

In any case, more rigorous tests would also help us try out other patches for the queues. For example, `Sql` uses table-locks for `claimItem` but not `createItem` or `deleteItem`... maybe they should all use table-locks? Or, maybe we don't need table-locks at all, if `claimItem` is rewritten to squeeze the SELECT+UPDATE into one query? Or, maybe each queue should be a separate table -- so that there is less shared indexing? All of those are plausible, and writing those patches isn't necessarily hard... but playing in that space requires a good "measure of success"...

totten:
Note: Most comments above are about a subthread/tangent.

For this PR, I guess the main review-question is... does it work as advertised? Anything else?

larssandergreen:
We've had this in production for a couple weeks and it has resolved the issue without any noticeable side effects. Can we get this into 6.19?

colemanw:
Yes this PR looks correct to me as well. Just noting that MariaDB and MySQL use the same InnoDB engine error codes,  so this will work for both.

larssandergreen:
Thanks @colemanw

@totten re: logs above: unfortunately, deadlocks in the logs are too uneven to say anything specific at this point. It looks like we've just had far fewer deadlocks in general over the last two weeks.

larssandergreen:
@totten I've [done a PR](https://github.com/civicrm/civicrm-core/pull/36635) to remove the table locks from queue_items. We are getting a fair number of deadlocks from this and I don't think we need the table-level locking.

PR #447, opened 2026-07-23
Remove obsolete folder, `_ORIGINAL_/Smarty`

totten:
The `_ORIGINAL_/` folder is left over from a painful procedure used to upgrade forked packages. In this case, Smarty v2.

But, now-a-days, CiviCRM uses a de-forked instance of Smarty v5.

That upgrade procedure -- and this folder -- is obsolete.


colemanw:
Good riddance!

